PR #118, opened 2023-04-06
Nathan/sad path integration tests part 2

nathanosdev:
I've added more sad path tests here but also added this `rstest` library: https://github.com/la10736/rstest
It supports fixture injection and parameterized tests. I've refactored _some_ tests, but not all, to keep the PR size smaller.
I'll continue to refactor more tests and follow up with more PRs.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -779,58 +779,93 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


keplervital (approved):
LGTM

PR #131, opened 2023-04-28
test: add JS e2e tests

nathanosdev:
This PR adds JS e2e tests for the response verification package.

### DIFF
--- a/packages/ic-response-verification-tests/Cargo.toml
+++ b/packages/ic-response-verification-tests/Cargo.toml
@@ -6,8 +6,8 @@ edition = "2021"
 [dependencies]
 hex = "0.4.3"
 http = "0.2.8"
-ic-agent = "0.23.1"
-ic-utils = "0.23.1"
+ic-agent = { git = "https://github.com/dfinity/agent-rs.git" }
+ic-utils = { git = "https://github.com/dfinity/agent-rs.git" }
 tokio = { version = "1.24.2", features = ["full"] }
 ic-response-verification = { path = "../ic-response-verification", features = ["debug"] }
 anyhow = "1.0.68"
--- a/packages/ic-response-verification-tests/src/agent.rs
+++ b/packages/ic-response-verification-tests/src/agent.rs
@@ -2,10 +2,11 @@ use anyhow::Result;
 use ic_agent::agent::http_transport::ReqwestHttpReplicaV2Transport;
 use ic_agent::Agent;
 
-pub fn create_agent(url: &str) -> Result<Agent> {
+pub async fn create_agent(url: &str) -> Result<Agent> {
     let transport = ReqwestHttpReplicaV2Transport::create(url)?;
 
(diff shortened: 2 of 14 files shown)


packages/ic-response-verification-tests/Cargo.toml
```diff
@@ -6,8 +6,8 @@ edition = "2021"
 [dependencies]
 hex = "0.4.3"
 http = "0.2.8"
-ic-agent = "0.23.1"
-ic-utils = "0.23.1"
+ic-agent = { git = "https://github.com/dfinity/agent-rs.git" }
+ic-utils = { git = "https://github.com/dfinity/agent-rs.git" }
```
nathanosdev, line 10: There was a bug with `ic-utils`. This is affecting the e2e tests and ICX Proxy.

https://github.com/dfinity/agent-rs/pull/432

packages/ic-response-verification-tests/src/agent.rs
```diff
@@ -2,10 +2,11 @@ use anyhow::Result;
 use ic_agent::agent::http_transport::ReqwestHttpReplicaV2Transport;
 use ic_agent::Agent;
 
-pub fn create_agent(url: &str) -> Result<Agent> {
+pub async fn create_agent(url: &str) -> Result<Agent> {
```
nathanosdev, line 5: I also updated the rust tests to use the root key from the agent instead of passing it in from the script.
Retrospectively this feels like the obvious thing to do 😆

packages/ic-response-verification-tests/wasm-tests/http-interface/canister_http_interface.did
```diff
@@ -0,0 +1,47 @@
+// adapted from https://internetcomputer.org/docs/current/references/ic-interface-spec/#http-gateway-interface
```
nathanosdev, line 1: I think it would be good to add a `http-canister` package that encapsulates this interface, like what we have in `ic-utils`. I'll add a ticket for this in our epic on breaking the service worker into smaller packages.

packages/ic-response-verification-tests/Cargo.toml
keplervital: Can we pin here to a specific commit hash?

keplervital (approved):

packages/ic-response-verification-tests/Cargo.toml
nathanosdev: Cargo is doing this automatically: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories
> Once a git dependency has been added, Cargo will lock that dependency to the latest commit at the time. New commits will not be pulled down automatically once the lock is in place. However, they can be pulled down manually with cargo update.

And I hope to update it this week anyway once `agent-rs` does another release.

PR #149, opened 2023-05-12
(refactor: TT-79) Standardize verification errors as reasons

keplervital:
Updated the verification error handling to include a reason when verification fails. This now means that all verification related issues are not thrown anymore, instead they are include in the `reason` property of the response.

The errors that are still thrown are now related to the parsing logic before the [verification](https://github.com/dfinity/response-verification/blob/kepler/tt-79-errors/packages/ic-response-verification/src/verification/verify_request_response_pair.rs#L83) method is called upon or the verification version is not supported (this case still results in the error `UnsupportedVerificationVersion`)

### DIFF
--- a/packages/ic-response-verification/tests/v2_response_verification_certification_scenarios.rs
+++ b/packages/ic-response-verification/tests/v2_response_verification_certification_scenarios.rs
@@ -183,6 +183,7 @@ mod tests {
             result,
             VerificationResult::Failed {
                 verification_version,
+                reason: _,
             } if verification_version == 2
         ));
     }
@@ -291,6 +292,7 @@ mod tests {
             result,
             VerificationResult::Failed {
                 verification_version,
+                reason: _
             } if verification_version == 2
         ));
     }
--- a/packages/ic-response-verification/tests/v2_response_verification_certification_scenarios.rs
+++ b/packages/ic-response-verification/tests/v2_response_verification_certification_scenarios.rs
@@ -12,7 +12,7 @@ mod tests {
     };
     use ic_response_verification::{
         types::{Request, Response, VerificationResult, VerifiedResponse},
-        verify_request_response_pair,
+        verify_request_response_pair, ResponseVerificationError,
     };
     use ic_response_verification_test_utils::{
         create_v2_certificate_fixture, create_v2_header, get_current_timestamp, ExprTree,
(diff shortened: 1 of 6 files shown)


packages/ic-response-verification/tests/v2_response_verification_certification_scenarios.rs
```diff
@@ -183,6 +183,7 @@ mod tests {
             result,
             VerificationResult::Failed {
                 verification_version,
+                reason: _,
```
nathanosdev, line 186: Could we add the expected reason as a parameter to the tests? Or do you think it would be overkill?

packages/ic-response-verification/tests/v2_response_verification_sad_path.rs
```diff
@@ -344,9 +348,13 @@ mod tests {
...
             &root_key,
             MIN_REQUESTED_VERIFICATION_VERSION,
-        );
+        )
+        .unwrap();
 
-        assert!(result.is_err());
+        assert!(matches!(result, VerificationResult::Failed {
+            verification_version,
+            reason: _,
```
nathanosdev, line 356: Same question here regarding adding the reason as a parameter

packages/ic-response-verification/tests/v2_response_verification_certification_scenarios.rs
keplervital: All errors here are `InvalidExpressionPath`, so, initially i thought it would be an overkill to repeat this for each test case, but since it's not a big effort i've just added either way, it will help us in the future if any of those case scenarios deviate from the expect error.

packages/ic-response-verification/tests/v2_response_verification_sad_path.rs
keplervital: Added 👍  for the `CertificateTimeTooFarInThePast` & `CertificateTimeTooFarInTheFuture` i've ignored the values and just checked if those are the failure reasons. Do you think it would add any value checking the exact times of those errors?

nathanosdev: That's good! We have unit tests to cover the exact times of those errors so I don't think we need to repeat it here 👍

nathanosdev (approved):

PR #152, opened 2023-05-26
fix: add missing request body hashing

nathanosdev:
Request body hashing is part of the spec, but has not been part of the implementation: https://github.com/dfinity/interface-spec/blob/master/spec/http-gateway-protocol-spec.md#request-hash-calculation

Thankfully nobody is using this yet 🙃

### DIFF
--- a/packages/ic-response-verification/src/hash/response_hash.rs
+++ b/packages/ic-response-verification/src/hash/response_hash.rs
@@ -120,7 +120,6 @@ pub fn response_hash(
     response: &Response,
     response_certification: &ResponseCertification,
 ) -> Sha256Digest {
-    // lower case the headers here
     let filtered_headers = filter_response_headers(response, response_certification);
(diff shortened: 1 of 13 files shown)


packages/ic-response-verification/src/hash/response_hash.rs
```diff
@@ -120,6 +120,7 @@ pub fn response_hash(
     response: &Response,
     response_certification: &ResponseCertification,
 ) -> Sha256Digest {
+    // lower case the headers here
```
keplervital, line 123: is this a todo?

keplervital (approved):

packages/ic-response-verification/src/hash/response_hash.rs
nathanosdev: Thanks for spotting that. I meant to remove this. it's done now and I created a ticket to follow up on that instead.

PR #154, opened 2023-05-31
test: do not build package twice for e2e tests and catch JS errors

nathanosdev:


scripts/e2e.sh
```diff
@@ -120,7 +120,6 @@ run_e2e_tests() {
 
 download_sdk_repo
 build_dfx
-./scripts/package.sh
```
nathanosdev, line 123: this is already being run in the `run_e2e_tests` function

keplervital (approved):
LGTM

PR #160, opened 2023-06-28
feat: add certificate verification JS library

nathanosdev:
re #TT-253


examples/certified-counter/package.json
```diff
@@ -0,0 +1,48 @@
...
+  ],
+  "scripts": {
+    "build": "webpack",
+    "prebuild": "npm run generate",
+    "start": "webpack serve --mode development --env development",
+    "prestart": "npm run generate",
+    "generate": "dfx generate backend"
+  },
+  "dependencies": {
+    "@dfinity/certificate-verification": "workspace:*"
```
keplervital, line 19: Since this `package.json` is using the `pnpm` format with the dependencies version we should also use `pnpm` on the scripts section imo

packages/certificate-verification-js/src/error.ts
```diff
@@ -0,0 +1,13 @@
+export class CertificateVerificationError extends Error {
+  constructor(message: string) {
+    super(message);
+    this.name = 'CertificateVerificationError';
```
keplervital, line 4: you should be able to use `this.constructor.name` in the errors.

keplervital (approved):
as discussed offline we should add some tests to this, but the library looks good

nathanosdev:
Tests are failing, they are fixed in this PR: https://github.com/dfinity/response-verification/pull/161

PR #161, opened 2023-06-30
test: add more integration tests

nathanosdev:


keplervital (approved):
LGTM

I've updated the v2 load assets logic to reuse the same function and receive file_path and http_path as arguments

PR #164, opened 2023-07-20
feat: add test utils wasm package

nathanosdev:
add unit test for certificate verification JS library

re #TT-296


nathanosdev:
I've added a single unit test for the certificate verification JS library to ensure that the library is really working. I'll follow up with another PR to add more unit tests and then also refactor the existing tests in the response verification library to use the new testing library.

examples/certified-counter/src/frontend/src/index.ts
```diff
@@ -33,13 +33,13 @@ buttonElement.addEventListener('click', async event => {
...
   const agent = new HttpAgent();
   await agent.fetchRootKey();
-  const tree = await verifyCertification(
-    Principal.fromText(canisterId),
-    new Uint8Array(certificate).buffer,
-    new Uint8Array(witness).buffer,
-    agent.rootKey,
-    50000,
-  );
+  const tree = await verifyCertification({
```
nathanosdev, line 36: Some feedback from David, this would be cleaner as an object rather than separate params. I've no strong opinion either way.

scripts/package.sh
```diff
@@ -1,26 +1,32 @@
 PKG_ROOT=$1
 OUT_DIR=$2
 
+early_exit() {
```
nathanosdev, line 4: While working on the library I had missed some Rust error, this allows the script to exit early instead of trying to run all 4 builds.

.github/actions/setup-dfx/action.yml
```diff
@@ -22,7 +18,7 @@ runs:
           echo "DFX restored from cache"
         else
           echo "DFX not restored from cache, running install script:"
-          DFX_VERSION=${{ env.dfx_version }} sh -ci "$(curl -fsSL https://sdk.dfinity.org/install.sh)"
+          DFX_VERSION=0.14.2 sh -ci "$(curl -fsSL https://sdk.dfinity.org/install.sh)"
```
nathanosdev, line 21: I had copy pasted this from another project. The other project had a root `dfx.json` but response verification doesn't. For now I think it's fine to lock this down manually. But once we have this feature in Cargo: https://github.com/rust-lang/cargo/issues/9406 we should be able to include the DFX projects into the Cargo workspace and then create a single `dfx.json` file to handle all DFX projects and clean this up.

rust-toolchain.toml
```diff
@@ -0,0 +1,4 @@
+[toolchain]
+channel = "1.70.0"
```
nathanosdev, line 2: The pipeline had an issue with the latest stable (`1.71.1`) so I added this to lock down the version. This was something on my todo list anyway.

packages/certificate-verification-js/src/index.ts
```diff
@@ -9,13 +9,21 @@ import { Principal } from '@dfinity/principal';
...
 import { CertificateTimeError, CertificateVerificationError } from './error';
 
-export async function verifyCertification(
-  canisterId: Principal,
-  encodedCertificate: ArrayBuffer,
-  encodedTree: ArrayBuffer,
-  rootKey: ArrayBuffer,
-  maxCertificateTimeOffsetMs: number,
-): Promise<HashTree> {
+export interface CertificationParams {
```
keplervital, line 12:
```suggestion
export interface VerifyCertificationParams {
```

If you wouldn't like to associate the name exactly to the function then you could still do `CertificationParams` but then have an alias like `type VerifyCertificationParams = CertificationParams`

nathanosdev: I think associating the name to the function makes sense since the interface is dedicated for it, thanks for the suggestion!

keplervital (approved):

PR #172, opened 2023-08-02
feat: always return a result/exception when response verification fails

nathanosdev:
This PR updates the return type to always be a result/exception, instead of only for "unexpected" scenarios.
There are also some Cargo clippy fixes and a fix for the e2e tests that allowed broken e2e tests to pass in a previous PR: https://github.com/dfinity/response-verification/actions/runs/5737416933/job/15549030697

### DIFF
--- a/packages/ic-response-verification-tests/wasm-tests/main.ts
+++ b/packages/ic-response-verification-tests/wasm-tests/main.ts
@@ -1,5 +1,5 @@
 import {
-  VerificationResult,
+  VerificationInfo,
   Request,
   Response,
   getMinVerificationVersion,
@@ -15,6 +15,7 @@ import {
 import { HttpAgent, ActorSubclass, Actor, Agent } from '@dfinity/agent';
 import { Principal } from '@dfinity/principal';
 import assert from 'node:assert';
+import { exit } from 'process';
 
 async function createAgentAndActor(
   gatewayUrl: string,
@@ -49,6 +50,7 @@ async function main(): Promise<void> {
     await v2Test(principal, agent, actor);
   } catch (error) {
     console.error('Error running e2e tests...', error);
+    exit(1);
   }
 }
 
@@ -65,7 +67,6 @@ async function v1Test(
     agent,
     actor,
   );
-  assert(resultOne.passed);
   assert.equal(resultOne.verificationVersion, 1);
 
   const resultTwo = await performTest(
@@ -76,7 +77,6 @@ async function v1Test(
     agent,
     actor,
   );
-  assert(resultTwo.passed);
   assert.equal(resultTwo.verificationVersion, 1);
 }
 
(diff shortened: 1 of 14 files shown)


packages/ic-response-verification-tests/wasm-tests/main.ts
```diff
@@ -49,6 +50,7 @@ async function main(): Promise<void> {
     await v2Test(principal, agent, actor);
   } catch (error) {
     console.error('Error running e2e tests...', error);
+    exit(1);
```
nathanosdev, line 53: If we encounter an error, make sure to exit with a non-zero exit code so that the CI script picks up on it.

packages/ic-response-verification/src/cbor/hash_tree.rs
```diff
@@ -23,21 +23,21 @@ pub fn parsed_cbor_to_tree(parsed_cbor: &CborValue) -> Result<HashTree, Response
         cbor_tags.reverse();
 
         if let Some(CborValue::HashTree(hash_tree_tag)) = cbor_tags.pop() {
-            return match hash_tree_tag {
+            match hash_tree_tag {
```
nathanosdev, line 26: Clippy fixes

packages/ic-response-verification/src/types/verification_result.rs
```diff
@@ -1,114 +1,62 @@
...
-#[cfg(all(target_arch = "wasm32", feature = "js"))]
-use crate::ResponseVerificationJsError;
-
 #[cfg(all(target_arch = "wasm32", feature = "js"))]
 #[wasm_bindgen(typescript_custom_section)]
 const VERIFICATION_RESULT: &'static str = r#"
-type VerificationResult = {
-  passed: true;
+type VerificationInfo = {
   response?: VerifiedResponse;
```
keplervital, line 10: Since we don't have `passed` anymore is there any situation where the response would not be set?

packages/ic-response-verification/src/types/verification_result.rs
```diff
@@ -1,114 +1,62 @@
...
-        /// The version of verification that was used to verify the response
-        verification_version: u16,
-        /// The reason why the verification of the response has failed
-        reason: ResponseVerificationError,
-    },
+pub struct VerificationInfo {
+    /// Response object including the status code, body and headers that were included in the
+    /// certification and passed verification. If verification failed then this object will be
+    /// empty.
+    pub response: Option<VerifiedResponse>,
```
keplervital, line 21: same here with the `Option`, ins't this throwing an exception now when verification fails?

packages/ic-response-verification/src/types/verification_result.rs
nathanosdev: Yes, you can still skip certification for a given path entirely with a CEL expression. In this case we return `None` for the `response`: https://github.com/dfinity/response-verification/blob/main/packages/ic-response-verification/src/verification/verify_request_response_pair.rs#L253-L261

keplervital: I think we are currently not handling that scenario then from the clients of response verification, (e.g. [service-worker returns a failed verification response](https://github.com/dfinity/ic/blob/205fc8bb8c82238f0a750e7e0464da864c8b3d39/typescript/service-worker/src/sw/requests/query-call.ts#L40))

keplervital (approved):

PR #191, opened 2023-08-16
feat: add ic-certification library

nathanosdev:
This moves the `ic-certification` library from https://github.com/dfinity/agent-rs.
I haven't made any code changes, although I made some slight alterations to the `Cargo.toml`. All changes outside of the `ic-certification` folder are the usual checklist items for adding new packages.


keplervital (approved):
LGTM

the only thing that might look like a strange change imo is the jump from version `0.25` to `1.0.2` since now this library will follow this repo versioning, but i don’t think we can do anything about it at this point if we want to keep the single versioning within the repo

PR #196, opened 2023-08-24
fix: url decoding for encoded query strings

keplervital:
re #TT-300

Query strings were being decoded the same way as the entire URL which would break the URI format, this fixes it by handling the path and query string differently.

### DIFF
--- a/packages/ic-response-verification/src/types/request.rs
+++ b/packages/ic-response-verification/src/types/request.rs
@@ -30,9 +30,33 @@ pub struct Request {
 
 impl Request {
     pub(crate) fn get_uri(&self) -> ResponseVerificationResult<Uri> {
-        let url = urlencoding::decode(&self.url)?;
+        let uri = self
+            .url
+            .parse::<Uri>()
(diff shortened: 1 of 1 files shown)


packages/ic-response-verification/src/types/request.rs
```diff
@@ -30,9 +30,33 @@ pub struct Request {
...
     pub(crate) fn get_uri(&self) -> ResponseVerificationResult<Uri> {
-        let url = urlencoding::decode(&self.url)?;
+        let uri = self
+            .url
+            .parse::<Uri>()
+            .map_err(|_| ResponseVerificationError::MalformedUrl(self.url.clone()))?;
+
+        // When decoding a URL the path and query string need to be handled differently, this allows for
+        // clients to send encoded URL paths that would match the same decoded URL path but the query string can
+        // still ramain encoded.
```
nathanosdev, line 40:
```suggestion
        // still remain encoded.
```

nathanosdev (approved):

PR #213, opened 2023-09-14
fix: handle non-latin characters in URL path

nathanosdev:
re #TT-319

### DIFF
--- a/packages/ic-response-verification/src/types/request.rs
+++ b/packages/ic-response-verification/src/types/request.rs
@@ -29,34 +29,20 @@ pub struct Request {
 }
 
 impl Request {
-    pub(crate) fn get_uri(&self) -> ResponseVerificationResult<Uri> {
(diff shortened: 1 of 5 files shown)


packages/ic-response-verification/src/types/request.rs
```diff
@@ -29,34 +29,20 @@ pub struct Request {
...
-            .query()
-            .map(|query| format!("?{}", query))
-            .unwrap_or_default();
-
-        let decoded_url = format!(
-            "{}{}{}{}",
-            scheme_with_delimiter, authority, decoded_path, query_string
-        );
+        let decoded_url = urlencoding::decode(uri.path()).map(|path| path.into_owned())?;
+        Ok(decoded_url)
```
keplervital, line 39:
```suggestion
        let decoded_path = urlencoding::decode(uri.path()).map(|path| path.into_owned())?;
        Ok(decoded_path)
```

keplervital (approved):

PR #217, opened 2023-10-02
fix: update dependencies

nathanosdev:


.github/workflows/release.yml
```diff
@@ -53,6 +53,11 @@ jobs:
         env:
           CRATES_TOKEN: ${{ secrets.CRATES_TOKEN }}
 
+      - name: Release ic-certification Cargo crate
```
nathanosdev, line 56: Need to release this before `ic-cbor`

examples/certified-counter/src/backend/Cargo.toml
```diff
@@ -15,4 +15,4 @@ ic-cdk-macros = "0.6.0"
 ic-certified-map = "0.3.4"
 serde = "1.0.164"
 serde_cbor = "0.11.2"
-sha256 = "1.1.4"
```
nathanosdev, line 18: For some reason this library depends on `tokio-rs`. After updating packages, `tokio-rs` introduced a breaking change for WASM compilation in a minor version bump :) +1 Rust version control standards are great!!!

PR #224, opened 2023-10-31
ci: publish certification-testing and certificate-verification libraries

nathanosdev:
We were asked about [publishing the ic-certification](https://dfinity.slack.com/archives/C03P8DZ2PTJ/p1698414327916929) library. Unfortunately we can't due to Git dependencies in the main IC repo, but I've added the step anyway with some comments explaining why it's not published yet. I also added some missing usage docs and the publish steps for the JS counterpart. I also added the step for `@dfinity/certificate-verification` because it was missing. `ic-certificate-verification` is already published.

### DIFF
--- a/packages/ic-certification-testing-wasm/README.md
+++ b/packages/ic-certification-testing-wasm/README.md
@@ -2,7 +2,7 @@
 
 [Certificate verification](https://internetcomputer.org/docs/current/references/ic-interface-spec#canister-signatures) on the [Internet Computer](https://dfinity.org) is the process of verifying that a canister's response to a [query call](https://internetcomputer.org/docs/current/references/ic-interface-spec#http-query) has gone through consensus with other replicas hosting the same canister.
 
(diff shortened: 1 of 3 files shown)


packages/ic-certification-testing-wasm/README.md
```diff
@@ -1 +1,59 @@
 # Certification Testing
+
+[Certificate verification](https://internetcomputer.org/docs/current/references/ic-interface-spec#canister-signatures) on the [Internet Computer](https://dfinity.org) is the process of verifying that a canister's response to a [query call](https://internetcomputer.org/docs/current/references/ic-interface-spec#http-query) has gone through consensus with other replicas hosting the same canister.
+
+This package provides a set of utilities to create these certificates for the purpose of testing in any Rust client that may need to verify them.
```
keplervital, line 5:
```suggestion
This package provides a set of utilities to create these certificates for the purpose of testing in any Javascript client with `wasm` support that may need to verify them.
```

keplervital (approved):

PR #228, opened 2023-11-16
chore: migrate wasm crates to workspace

nathanosdev:
This moves all of the crates to the root workspace.
There's a workaround for the WASM crates by using the `default-members` of the root workspace. It's not ideal, but while we wait for https://github.com/rust-lang/cargo/issues/9406 I think this is a better situation. I'll also follow up with a root DFX workspace now, currently there's two separate ones and likely more to come as more examples are added for the certification libraries so it will be nice to have that too.


.vscode/settings.json
```diff
@@ -0,0 +1,6 @@
+{
```
nathanosdev, line 1: Small, unrelated change. This causes VSCode to build in a separate directory from `cargo build` so they don't compete for file locks.

Cargo.toml
```diff
@@ -10,15 +10,109 @@ members = [
...
+js-sys = "0.3"
+console_error_panic_hook = "0.1"
+wasm-bindgen-console-logger = "0.1"
+# https://github.com/rust-random/rand#wasm-support
+# https://docs.rs/getrandom/latest/getrandom/#webassembly-support
+rand = "0.8"
+getrandom = { version = "0.2", features = ["js"] }
+
+
+ic-certification = { path = "./packages/ic-certification", default-features = false, version = "1.2.0" }
```
nathanosdev, line 76: Unfortunately I couldn't find a way to not manually specify these versions. Without them Cargo complains when trying to publish and I can't figure out how to reference `workspace.package.version` here :(

keplervital: For those local crates that have their own `Cargo.toml` ins't cargo taking that version if you omit here? 🤔

keplervital (approved):

Cargo.toml
nathanosdev: If `ic-certification` specifies it's version as `workspace.version` then that's the version it will take when publishing to crates.io, but it will not read the version of `ic-certification`'s dependencies from their `Cargo.toml`.

For example, when publishing `ic-response-verification` which depends on `ic-certification`, it will read `ic-response-verification`'s `Cargo.toml` but not `ic-certification`'s `Cargo.toml` when publishing. That's why the version also needs to be specified manually for the dependencies.

PR #234, opened 2023-11-24
feat: add-ic-http-certification-crate

nathanosdev:
add CEL expression generation

This PR is 80% documentation, 19% tests and 1% code 😅


.github/workflows/create-release-pr.yml
```diff
@@ -34,6 +34,10 @@ jobs:
           push: false
           github_token: ${{ secrets.GITHUB_TOKEN }}
 
+      # commitizen is formatting the .cz.yaml file in a way that Prettier does not like
```
nathanosdev, line 37: This is something I've had to manually fix on the last few releases.

packages/ic-http-certification/src/lib.rs
```diff
@@ -0,0 +1,242 @@
+//! # Internet Computer HTTP Certification
```
nathanosdev, line 1: Unfortunately this is the same as `README.md`, but with links added to items so it will navigate around docs.rs in a nice way. I'm a bit conflicted by duplicating this, but it ensures that everywhere a developer encounters the library, they will have documentation to greet them. docs.rs doesn't have an option to include a readme :(

keplervital (approved):
LGTM

.github/workflows/create-release-pr.yml
nathanosdev: Done, thanks for the suggestion!

PR #239, opened 2023-11-30
feat: add copy on write support to CEL definition types

nathanosdev:
When integrating `ic-http-certification` into `ic-response-verification`, I ran into issues with managing lifetimes because the `DefaultRequestCertification` and `DefaultResponseCertification` structs are holding references to data, instead of owning it.

This causes problems when you want to return a reference to data that is owned by the current function. Using copy on write ([COW](https://doc.rust-lang.org/std/borrow/enum.Cow.html)) for the properties on these structs allows us to keep the advantage of holding references to data while allowing the struct to take ownership of data in those cases where it's needed.

I split this PR out from the changes in `ic-response-verification` to make the PR smaller.

### DIFF
--- a/packages/ic-http-certification/src/cel/cel_types.rs
+++ b/packages/ic-http-certification/src/cel/cel_types.rs
@@ -1,4 +1,5 @@
 use super::create_cel_expr;
+use std::borrow::Cow;
 
 /// A certification CEL expression defintion.
 /// Contains an enum variant for each CEL function supported for certification.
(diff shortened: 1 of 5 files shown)


packages/ic-http-certification/src/cel/cel_types.rs
```diff
@@ -75,17 +76,35 @@ pub enum DefaultResponseCertification<'a> {
...
 
     /// A list of response headers to exclude from certification.
     ///
     /// As many or as little headers can be provided as desired.
     /// Providing an empty list will result in all response headers being certified.
-    ResponseHeaderExclusions(&'a [&'a str]),
+    ResponseHeaderExclusions(Cow<'a, [&'a str]>),
+}
+
+impl<'a> DefaultResponseCertification<'a> {
```
nathanosdev, line 88: I added these convenience methods so developers don't have to write `DefaultResponseCertification::CertifiedResponseHeaders(Cow::Borrowed(&headers))` and they can just write `DefaultResponseCertification::certified_response_headers(&headers)`.

PR #243, opened 2023-12-05
fix: missing certification object in CEL expressions

nathanosdev:
After finishing the integration with the `ic-response-verification` the tests started failing because the CEL expressions were not valid 🙃


keplervital (approved):
LGTM

PR #245, opened 2023-12-05
feat: move request and response hashing from ic-response-verification to ic-http-certification

nathanosdev:


packages/ic-response-verification/tests/v2_response_verification_certification_scenarios.rs
```diff
@@ -296,13 +296,12 @@ mod tests {
```
nathanosdev, line 296: Using the CEL builder in our integration tests means we have a lot less handwritten CEL expressions. Eventually I'd like to get rid of all of them except for the ones that are used to test the CEL parser and builder.

There's still a few more around, but refactoring the tests wasn't my focus. I ran into issues in these tests because of changes that are coming in another PR. The Certification object from ic-http-certification is using references now, so after integrating with that the CEL parser functions have some lifetime restrictions on their parameters that makes them difficult to work with in certain situations and it was much easier to just use the CEL builder in the tests.

packages/ic-http-certification/src/hash/request_hash.rs
```diff
@@ -37,14 +35,14 @@ pub fn request_hash(
...
-    request_certification: &RequestCertification,
+    request_certification: &DefaultRequestCertification,
 ) -> Vec<(String, Value)> {
     headers
         .iter()
         .filter_map(|(header_name, header_value)| {
             let is_header_included =
                 request_certification
-                    .certified_request_headers
+                    .headers
```
keplervital, line 45: any reason for the change here, wouldn't this now include headers that shouldn't go through certification?

packages/ic-http-certification/src/hash/request_hash.rs
```diff
@@ -71,11 +69,12 @@ fn get_filtered_query(query: &str, request_certification: &RequestCertification)
 
             query_param_name
                 .map(|query_param_name| {
-                    request_certification.certified_query_parameters.iter().any(
-                        |query_param_to_include| {
+                    request_certification
+                        .query_parameters
```
keplervital, line 73: same here as above

keplervital (approved):

PR #260, opened 2024-01-03
feat(ic-http-certification): add certification builder

nathanosdev:
improve underlying cel structures to prevent invalid cel expressions from being created


keplervital (approved):
LGTM

PR #262, opened 2024-01-10
feat: add bls signature cache

nathanosdev:


packages/ic-certificate-verification/src/signature_verification/reproducible_rng.rs
nathanosdev, line 1: Taken from https://github.com/dfinity/ic/blob/master/rs/crypto/test_utils/reproducible_rng/src/lib.rs with some code removed that we don't need.

packages/ic-certificate-verification/src/signature_verification/tests.rs
nathanosdev, line 1: Taken from https://github.com/dfinity/ic/blob/master/rs/crypto/internal/crypto_lib/threshold_sig/bls12_381/src/cache/tests.rs/. No changes aside from imports.

packages/ic-certificate-verification/src/signature_verification/signature_cache.rs
nathanosdev, line 1: Taken from https://github.com/dfinity/ic/blob/master/rs/crypto/internal/crypto_lib/threshold_sig/bls12_381/src/cache.rs. There's a minor change to interface from sized slices to non-sized.

packages/ic-certificate-verification/src/signature_verification/mod.rs
```diff
@@ -0,0 +1,32 @@
...
+
+pub fn verify_signature(
+    pk: &[u8],
+    sig: &[u8],
+    msg: &[u8],
+) -> Result<(), CertificateVerificationError> {
+    let entry = SignatureCacheEntry::new(pk, sig, msg);
+
+    if SignatureCache::global().contains(&entry) {
+        return Ok(());
```
altkdf, line 21: Note: it could be useful to somehow expose the cache hits/misses for metrics to see how many verifications are costly.

altkdf:
LGTM

packages/ic-certificate-verification/src/signature_verification/mod.rs
nathanosdev: Thanks, yes I discussed this with @r-birkner and we agreed we wanted to take the shorter path and get this released ASAP. There's a redesign planned for ICX Proxy and that will be a good chance to rethink how we are handling metrics.

keplervital (approved):
LGTM

packages/ic-certificate-verification/src/signature_verification/reproducible_rng.rs
ninegua: Maybe I'm missing something, but what is the use of this, it doesn't appear to be used anywhere except in test?

ninegua: Oh I see, it is imported only under #[cfg(test)]. Makes sense!

PR #265, opened 2024-01-10
feat(ic-http-certification): add upgrade property to HttpResponse struct

nathanosdev:
Add the `upgrade` property on the `HttpResponse` object. Also rename `Certification` to `HttpCertification`.


packages/ic-cbor/src/cbor_parser.rs
```diff
@@ -246,7 +246,7 @@ pub fn parse_cbor_string_array(i: &[u8]) -> CborResult<Vec<String>> {
         .map(|elem| {
             let CborValue::ByteString(elem) = elem else {
                 return Err(CborError::UnexpectedCborNodeType {
-                    expected_type: "Array".into(),
+                    expected_type: "ByteString".into(),
```
nathanosdev, line 249: This should've had this value all along. It's inconsequential, but it helps with debugging.

keplervital (approved):
LGTM

PR #272, opened 2024-01-13
chore: bump ic repo

chenyan-dfinity:
`wasm-bindgen` is making breaking changes without bumping their major version. This PR also tries to fix that.

PR #273, opened 2024-01-15
chore: update ic repo references

nathanosdev:
Commits pulled from https://github.com/dfinity/response-verification/pull/272 to rewrite the commit messages to be commitizen compliant.


packages/ic-certification-testing/src/error.rs
```diff
@@ -55,10 +55,4 @@ impl From<serde_wasm_bindgen::Error> for CertificationTestError {
     }
 }
 
-impl From<CertificationTestError> for JsValue {
```
keplervital, line 58: i assume this was not in use right?

keplervital (approved):

PR #279, opened 2024-01-15
docs(ic-http-certification): add HTTP certified JSON api example

nathanosdev:


examples/http-certification/json-api/dfx.json
```diff
@@ -0,0 +1,11 @@
+{
+  "dfx": "0.15.2",
```
keplervital, line 2:
```suggestion
  "dfx": "0.15.3",
```

keplervital (approved):
LGTM

examples/http-certification/json-api/dfx.json
nathanosdev: Thanks! I updated it also in some other places.

PR #285, opened 2024-01-23
test(ic-http-certification): integrate ic-http-certification with ic-…

nathanosdev:
…repsonse-verification tests


packages/ic-http-certification/src/tree/certification_tree.rs
```diff
@@ -1,3 +1,5 @@
+use std::borrow::Borrow;
+
```
keplervital, line 2:
```suggestion
```

packages/ic-response-verification/tests/v2_response_verification_certification_scenarios.rs
```diff
@@ -346,6 +374,18 @@ mod fixtures {
         }
     }
 
+    #[fixture]
+    pub fn index_js_certification() -> HttpCertificationTreeEntry<'static> {
+        HttpCertificationTreeEntry {
```
keplervital, line 379: should we make it consistent with how this is being used above? `HttpCertificationTreeEntry::new`

keplervital (approved):

packages/ic-response-verification/tests/v2_response_verification_certification_scenarios.rs
nathanosdev: The issue here is on line 381. Since this is being returned from a function, we need to take the `HttpCertification` struct by value or Rust is complaining that the value doesn't live long enough 👴

packages/ic-http-certification/src/tree/certification_tree.rs
nathanosdev: Thanks I missed that!

PR #286, opened 2024-01-23
style: fix clippy issues

nathanosdev:


.github/workflows/build-and-test.yml
```diff
@@ -30,7 +30,7 @@ jobs:
         run: cargo test
 
       - name: Lint Rust
-        run: cargo clippy
+        run: cargo clippy --all-targets --all-features
```
nathanosdev, line 33: Make sure we run all targets so lints don't get missed.

packages/ic-cbor/src/lib.rs
```diff
@@ -1,3 +1,5 @@
+#![deny(clippy::all)]
```
nathanosdev, line 1: Make sure lints will actually fail the build.

keplervital (approved):
LGTM

PR #293, opened 2024-02-05
refactor: restructure examples and http certification tests

nathanosdev:
Example projects are now organized into folders according to the library that they serve as examples for:
```
examples/
  response-verification/
    nodejs/
    web/
    rust/
  certification/
    certified-counter/
  http-certification/
    json-api/
    assets/
```

The `response-verification` integration tests have been moved into a standalone crate since they also cover the `http-certification` lib.


keplervital (approved):
LGTM

PR #301, opened 2024-02-13
feat(ic-asset-certification): add initial asset router

nathanosdev:
This PR adds the `ic-asset-certification` crate with an initial, basic implementation of an asset router.
It supports registering assets by path or by glob patterns to the router, certifying user provided headers and basic fallback support.

I'd added docs for everything from the get go, including an example project and a user guide.

I also had `cargo doc` shouting at me because of some linting issues in other projects, so I fixed them and added an extra step to the CI pipeline to avoid a build up of issues in the docs again.


examples/http-certification/assets/src/backend/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+}
+
+#[query]
+fn http_request(req: HttpRequest) -> HttpResponse {
+    serve_asset(&req)
+}
+
+thread_local! {
+    static HTTP_TREE: RefCell<HttpCertificationTree> = RefCell::new(HttpCertificationTree::default());
+    static ASSET_ROUTER: RefCell<AssetRouter<'static>> = RefCell::new(AssetRouter::default());
```
keplervital, line 32: if you wrap the `HTTP_TREE` in a `Rc` you should be able to add it to the router with a builder pattern, which would be nice imo to avoid doing `HTTP_TREE.with` & `ASSET_ROUTER.with`

packages/ic-asset-certification/LICENSE
```diff
@@ -0,0 +1,202 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright 2023 DFINITY Foundation
```
keplervital, line 190:
```suggestion
   Copyright 2024 DFINITY Foundation
```

packages/ic-asset-certification/src/asset_router.rs
```diff
@@ -0,0 +1,719 @@
...
+            tree_entry,
+        }) = self.responses.get(&request.url)
+        {
+            let witness = http_certification_tree.witness(tree_entry, &request.url)?;
+            let expr_path = tree_entry.path.to_expr_path();
+
+            return Ok((response.clone().into(), witness, expr_path));
+        }
+
+        let mut url_scopes = request.url.split('/').collect::<Vec<_>>();
```
keplervital, line 136: i would add a comment here to note that the request.url will actually just include the path and not the domain part of the url, otherwise this scope parsing will also include the protocol part which is not what is intended, maybe also add a check for that since it's not enforced in the HttpRequest type

keplervital (approved):

packages/ic-asset-certification/LICENSE
nathanosdev: Thanks 😅

packages/ic-asset-certification/src/asset_router.rs
nathanosdev: That's a good point, we're assuming the request is coming from a correct implementation of the HTTP Gateway protocol. I think adding the validation, just in case, wouldn't be a bad thing. Or adding some more flexible parsing. I've added a ticket for that: https://dfinity.atlassian.net/browse/TT-395

examples/http-certification/assets/src/backend/src/lib.rs
nathanosdev: I agree. I'll give it a try and follow up with another PR

PR #302, opened 2024-02-13
feat: hide enums from interfaces that provide factories/constructors

nathanosdev:
When writing `DefaultResponseCerification::certified_response_headers` I would very often mistakenly write `DefaultResponseCerification::CertifiedResponseHeaders` instead and get an error because I was using it wrong.

The enum variants are not useful from outside the crate, so I've hidden them inside of a struct now and only the factories/constructors are exposed now. This will prevent other developers from getting tripped up by that same issue.

Previously you would use the enum variants if you had an owned value and the factory if you had a reference, but now I've adjusted the factories to accept both by using `impl Into<Cow>`.

I also added some additional `Cow`s in places where I realized I need them while working on the asset router.


.github/workflows/build-and-test.yml
```diff
@@ -27,7 +27,7 @@ jobs:
         run: cargo build --release
 
       - name: Test Cargo crates
-        run: cargo test
+        run: cargo test --all-features
```
nathanosdev, line 30: I had originally added an optional `assets` feature to the `ic-http-certification` crate and then needed this flag to run tests that were behind the feature flag. That's not needed anymore but I think it's good to keep this anyway.

examples/http-certification/custom-assets/README.md
```diff
@@ -333,9 +333,11 @@ Lastly, a function specifically to certify the `index.html` file. Since the fron
 This will allow the canister to return this file for any path that does not exactly match an existing path in the tree. If the canister tries to return this file instead of an exact match that exists, verification will fail.
 
 ```rust
-const INDEX_REQ_PATH: &str = "";
-const INDEX_TREE_PATH: HttpCertificationPath = HttpCertificationPath::Wildcard(INDEX_REQ_PATH);
-const INDEX_FILE_PATH: &str = "index.html";
+lazy_static! {
```
nathanosdev, line 336: Since we're not directly using the enum anymore and we're using factories we can't use `const`

examples/http-certification/custom-assets/README.md
```diff
@@ -347,14 +349,14 @@ fn certify_index_asset() {
...
-        .get_file(INDEX_FILE_PATH)
+        .get_file(*INDEX_FILE_PATH)
         .expect("No index.html file found!!!");
     let body = identity_file.contents();
 
     certify_asset(
         body,
         INDEX_FILE_PATH.to_string(),
-        INDEX_TREE_PATH,
+        &*INDEX_TREE_PATH,
```
keplervital, line 359:
```suggestion
        INDEX_TREE_PATH.to_owned(),
```

so that it's consistent to how this is being handled below

keplervital (approved):
LGTM

PR #310, opened 2024-02-23
feat: add support for hashing arrays of Value-objects

przydatek:
The current implementation is missing support for `Array`-values, this PR adds this support.


nathanosdev (approved):
Looks great! Thanks

PR #312, opened 2024-02-27
test(ic-http-certification): fix integration tests missing from workspace

nathanosdev:
The `ic-http-certification-tests` crate was not part of the default members of the workspace, so the tests were not run with `cargo test`. Some changes were missed because of that, but thankfully nothing major.

I also found some parts of the docs that were missed for the same changes.


keplervital (approved):
LGTM

PR #319, opened 2024-03-26
docs: security fixes for JSON API example project

nathanosdev:


packages/ic-response-verification/src/error.rs
nathanosdev, line 1: I added some additional error types for the CEL expression validation. This is code that I end up debugging quite frequently, so adding scenario specific errors helps to avoid the need to debug anything. I intend to add more for other parts of the code that may not be clear when they fail.

examples/http-certification/json-api/src/tests/jest.config.ts
nathanosdev, line 1: Added integration tests for the example project

examples/http-certification/json-api/src/backend/src/lib.rs
nathanosdev, line 1: The main security fixes are in this file. Adding additional functionality to "complete" the API was informational, but I think it serves as a better example project now.

packages/ic-http-certification-tests/tests/v2_response_verification_sad_path.rs
```diff
@@ -221,81 +220,7 @@ mod tests {
 
         assert!(matches!(
             result,
-            Err(ResponseVerificationError::InvalidExpressionPath)
-        ));
-    }
-
-    #[rstest]
```
nathanosdev, line 228: These tests are actually duplicates of the tests that were updated in the previous file so I just removed them.

examples/http-certification/json-api/src/backend/backend.did
```diff
@@ -8,13 +8,69 @@ type HttpRequest = record {
...
+    url : text;
+    headers : vec HeaderField;
+    body : blob;
+};
+
 type HttpResponse = record {
     status_code : nat16;
     headers : vec HeaderField;
     body : blob;
+    upgrade : opt bool;
```
keplervital, line 22: if we have a custom type above for the request that drops the `certificate_version` for update calls, i would also have another type for the response that drops the `upgrade`, otherwise i would keep in the request as well.

keplervital (approved):

examples/http-certification/json-api/src/backend/backend.did
nathanosdev: Good point. I opted to just drop it for now because I realized that we don't have the support on the Rust side for those types, so I added a ticket to handle that: https://dfinity.atlassian.net/browse/TT-392

PR #322, opened 2024-04-24
fix(ic-http-certification): fixed an issue where an asset that has a wildcard but not an exact path is not witnessed correctly

nathanosdev:
The issue I fixed here happens if you have an asset under a wildcard path, but not an exact path and then the asset is requested with a path that exactly matches the wildcard path (excluding the `<*>` postfix).

For example:
- `index.html` is added under the `/` wildcard
- `index.html` is added under the `index.html` exact path
- `index.html` is NOT added under the `/` exact path
- the `/` exact path is requested, which triggers a response with the `/` wildcard and an incorrect witness

If `index.html` is added under the `/` exact path then this issue wouldn't happen because we'd trigger the flow for exact match instead of a wildcard match.

### DIFF
--- a/packages/ic-http-certification/src/tree/certification_tree.rs
+++ b/packages/ic-http-certification/src/tree/certification_tree.rs
@@ -77,7 +77,10 @@ impl HttpCertificationTree {
 
                 // For wildcards we start by witnessing the path that was requested so the verifier can be sure
                 // that there is no exact match in the tree.
-                let requested_witness = self.tree.witness(&requested_tree_path);
+                let witness = self.tree.witness(&requested_tree_path);
+
+                // Then we witness the wildcard path that we will be responding with.
+                let witness = merge_hash_trees(witness, self.tree.witness(&responding_tree_path));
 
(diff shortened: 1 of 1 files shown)


packages/ic-http-certification/src/tree/certification_tree.rs
```diff
@@ -77,7 +77,10 @@ impl HttpCertificationTree {
 
                 // For wildcards we start by witnessing the path that was requested so the verifier can be sure
                 // that there is no exact match in the tree.
-                let requested_witness = self.tree.witness(&requested_tree_path);
+                let witness = self.tree.witness(&requested_tree_path);
+
+                // Then we witness the wildcard path that we will be responding with.
+                let witness = merge_hash_trees(witness, self.tree.witness(&responding_tree_path));
```
nathanosdev, line 83: The logic breaks on line 88. When the request path and the responding path have the same length, the loop never runs.
In the example the request path is `/` and the responding path is `/` (because we hit the wildcard).

keplervital (approved):
nice catch

PR #335, opened 2024-05-13
fix(ic-http-certification): ensure all possible wildcards are witnessed

nathanosdev:
This fixes an issue where wildcard paths that fully match the request URL were not witnessed.

For example, if you request `["foo", "bar", "baz"]` and respond with `["foo", "bar", "<*>"]` then you will be missing the witness for `["foo", "var", "baz", "<*>"]`.

This logic can actually be shared with similar code from `ic-response-verification`, so rather than changing the current code I refactored the `ic-response-verification` implementation to a common function.

### DIFF
--- a/packages/ic-http-certification/src/tree/certification_tree.rs
+++ b/packages/ic-http-certification/src/tree/certification_tree.rs
@@ -1,9 +1,10 @@
 use super::{
     certification_tree_entry::HttpCertificationTreeEntry,
-    certification_tree_path::{CertificationTreePathSegment, PATH_PREFIX_BYTES},
+    certification_tree_path::CertificationTreePathSegment,
 };
 use crate::{
-    tree::{HttpCertificationPathType, WILDCARD_PATH_TERMINATOR_BYTES},
+    tree::HttpCertificationPathType,
+    utils::{more_specific_wildcards_for, PATH_PREFIX_BYTES},
     HttpCertificationError, HttpCertificationPath, HttpCertificationResult,
 };
 use ic_certification::{labeled, labeled_hash, merge_hash_trees, AsHashTree, HashTree, NestedTree};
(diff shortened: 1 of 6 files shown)


packages/ic-http-certification/src/tree/certification_tree.rs
```diff
@@ -195,6 +179,10 @@ mod tests {
             not_found_witness.lookup_subtree(["http_expr", "assets", "js", "<*>"]),
             SubtreeLookupResult::Found(_)
         ));
+        assert_eq!(
+            not_found_witness.lookup_subtree(["http_expr", "assets", "js", "0.js", "<*>"]),
+            SubtreeLookupResult::Absent
```
nathanosdev, line 184: This covers the missing wildcard

packages/ic-response-verification/src/validation/v2_validation.rs
```diff
@@ -106,67 +91,46 @@ pub fn validate_expr_path(
...
                 request_path: request_path.to_string(),
             },
         );
     }
     request_url_path.pop(); // pop "<$>"
 
-    // if the expr_path matches the full URL with a wildcard
-    // there can't be a more precise path in the tree
-    request_url_path.push("<*>".into());
-    if original_path.eq(&request_url_path) {
```
nathanosdev, line 120: This code wasn't necessary, because we are anyway checking the full URL with a wildcard inside the loop

packages/ic-response-verification/src/validation/v2_validation.rs
```diff
@@ -2,45 +2,27 @@ use crate::{ResponseVerificationError, ResponseVerificationResult};
...
 use ic_certification::{hash_tree::Hash, HashTree, Label, SubtreeLookupResult};
 use ic_http_certification::cel::DefaultCelExpression;
+use ic_http_certification::utils::{
+    is_wildcard_path_valid_for_request_path, more_specific_wildcards_for,
+    EXACT_PATH_TERMINATOR_BYTES, PATH_PREFIX_BYTES,
+};
 use ic_http_certification::CelExpression;
 
-fn path_from_parts<T>(parts: &[T]) -> Vec<Label>
+fn path_from_parts<T>(parts: &[T]) -> Vec<Vec<u8>>
```
nathanosdev, line 11: There's a few changes here that are related to switching from `Label` to `Vec<u8>`. `Label` didn't have any advantage and `Vec<u8>` made it easier to share code with `ic-http-certification`.

packages/ic-response-verification/src/validation/v2_validation.rs
```diff
@@ -2,45 +2,27 @@ use crate::{ResponseVerificationError, ResponseVerificationResult};
...
+fn path_might_exist_in_tree(path: &[Vec<u8>], tree: &HashTree) -> bool {
     !matches!(tree.lookup_subtree(path), SubtreeLookupResult::Absent)
 }
 
-fn path_exists_in_tree(path: &[Label], tree: &HashTree) -> bool {
+fn path_exists_in_tree(path: &[Vec<u8>], tree: &HashTree) -> bool {
     matches!(tree.lookup_subtree(path), SubtreeLookupResult::Found(_))
 }
 
-fn is_wildcard_path_valid_for_request_path(
```
nathanosdev, line 26: Also common code now

keplervital (approved):

PR #341, opened 2024-06-06
feat(ic-asset-certification): add support for alt response encodings

nathanosdev:
This is an initial PR for adding support for alternative response encodings, Gzip, brotli etc...
It works and is fully tested, but I'm not entirely happy with the interface. It currently requires adding additional configuration entries for the encoded versions, and I'd like to improve that but I still haven't fully decided how a better interface will look.
In the meantime it would be nice to merge it in it's current state and then re-iterate on the interface in a future PR.

### DIFF
--- a/examples/http-certification/assets/src/backend/src/lib.rs
+++ b/examples/http-certification/assets/src/backend/src/lib.rs
@@ -1,7 +1,7 @@
 use base64::engine::general_purpose::STANDARD as BASE64;
 use base64::Engine;
 use ic_asset_certification::{
-    Asset, AssetConfig, AssetFallbackConfig, AssetRedirectKind, AssetRouter,
+    Asset, AssetConfig, AssetEncoding, AssetFallbackConfig, AssetRedirectKind, AssetRouter,
 };
 use ic_cdk::{
     api::{data_certificate, set_certified_data},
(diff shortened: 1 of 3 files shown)


examples/http-certification/assets/src/backend/src/lib.rs
```diff
@@ -47,7 +47,29 @@ fn collect_assets<'content, 'path>(
     assets: &mut Vec<Asset<'content, 'path>>,
 ) {
     for file in dir.files() {
-        assets.push(Asset::new(file.path().to_string_lossy(), file.contents()));
+        let file_path = file.path().to_string_lossy().to_string();
+
+        if file_path.ends_with(".br") {
```
nathanosdev, line 52: Ideally this extra code wouldn't be necessary

examples/http-certification/assets/src/backend/src/lib.rs
```diff
@@ -71,6 +93,30 @@ fn certify_all_assets() {
             }],
             aliased_by: vec!["/".to_string()],
         },
+        AssetConfig::File {
```
nathanosdev, line 96: Ideally also these extra configuration entries would not be necessary.

examples/http-certification/assets/src/backend/src/lib.rs
```diff
@@ -71,6 +93,30 @@ fn certify_all_assets() {
...
+                "cache-control".to_string(),
+                "public, no-cache, no-store".to_string(),
+            )]),
+            fallback_for: vec![AssetFallbackConfig {
+                scope: "/".to_string(),
+            }],
+            aliased_by: vec!["/".to_string()],
+        },
+        AssetConfig::File {
+            path: "index.html.br".to_string(),
```
keplervital, line 109: one alternative could be to have path be an enum that is either a specific file because it's only provided for one encoding or many, something like:

```
// for a single encoding
AssetConfig::File {
   path: AssetPath::File("index.html".to_string())
}

// for many encoding
AssetConfig::File {
   path: AssetPath::Alternate(vec![
     AssetFileEncoded::new("index.html".to_string(), Encoding::Identity)),
     AssetFileEncoded::new("index.html.gzip".to_string(), Encoding::Gzip)),
     AssetFileEncoded::new("index.html.br".to_string(), Encoding::Brotli)
   ])
}
```

or if you would want to make it strongly typed to avoid misconfigurations of adding to many times the same encoding:

```
AssetConfig::File {
   path: AssetPath::new("index.html".to_string())
     .with_gzip("index.html.gzip".to_string())
     .with_brotli("index.html.br".to_string())
     .build()
}
```

keplervital (approved):

PR #344, opened 2024-07-15
feat(ic-asset-certification): more ergonomic config for alternative asset encodings

nathanosdev:


examples/http-certification/assets/src/backend/src/lib.rs
```diff
@@ -80,6 +58,11 @@ fn collect_assets<'content, 'path>(
 // Certification
 fn certify_all_assets() {
     // 1. Define the asset certification configurations.
+    let encodings = vec![
+        (AssetEncoding::Brotli, "br".to_string()),
+        (AssetEncoding::Gzip, "gzip".to_string()),
+    ];
```
keplervital, line 64: maybe we could have an impl of `AssetEncoding` that returns that tuple e.g. 

```suggestion
    let encodings = vec![AssetEncoding::config_brotli(), AssetEncoding::config_gzip()];
```

packages/ic-asset-certification/README.md
```diff
@@ -284,34 +317,7 @@ let config = AssetConfig::Redirect {
...
-        scope: "/".to_string(),
-    }],
-  aliased_by: vec!["/".to_string()],
-};
-
-asset_router.certify_asset(asset, Some(asset_config)).unwrap();
-```
-
-Or in bulk using the `certify_assets` method:
+Assets can be inserted in bulk using the `certify_assets` method:
```
przydatek, line 320: nit: "in bulk" seems obsolete now

examples/http-certification/assets/src/backend/src/lib.rs
przydatek: Agreed, having some impl to avoid repetition would be good.
Maybe something like below would be better than a dedicated function for each encoding?
```
impl AssetEncoding {
  pub fn config_tuple(&self) -> (AssetEncoding, String) {
  ...
  }
}
```

przydatek (commented):
Looks good in general, please see some minor comments below.

packages/ic-asset-certification/src/asset_config.rs
```diff
@@ -217,6 +234,37 @@ pub enum AssetConfig {
...
+        /// the encoding of an asset, and the file extension for the encoded
+        /// asset. The router will search for an asset with the provided file
+        /// extension and certify all matching encoded assets, if found.
+        ///
+        /// A list of alternative encodings that can be used to serve the asset.
+        ///
+        /// Each entry is a tuple of the [encoding name](AssetEncoding) and the
+        /// file extension used in the file path. For example, to include Brotli
+        /// and Gzip encodings:
+        /// `vec![(AssetEncoding::Brotli, "br".to_string()), (AssetEncoding::Gzip, "gz".to_string())]`
```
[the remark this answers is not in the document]
nathanosdev: These lines are repeated on purpose because there is the same property repeated on two different variants.

```rust
pub enum AssetConfig {
  File {
    // ...
    encodings: Vec<(AssetEncoding, String)>,
  },
  Pattern {
    // ...
    encodings: Vec<(AssetEncoding, String)>,
  }
}
```

It's not ideal to duplicate it, but I think it's nice for developers to have the same information in both places so they don't have to look at docs of a different variant to understand.

examples/http-certification/assets/src/backend/src/lib.rs
nathanosdev: Thanks for the suggestion both.

I've added a slightly different approach like this:
```Rust
vec![AssetEncoding::Brotli.default(), AssetEncoding::Gzip.custom("gzip")]
```

What do you guys think?

packages/ic-asset-certification/src/asset_config.rs
```diff
@@ -324,6 +403,97 @@ pub enum AssetRedirectKind {
...
+    ///
+    /// let (encoding, extension) = AssetEncoding::Gzip.default();
+    /// assert_eq!(encoding, AssetEncoding::Gzip);
+    /// assert_eq!(extension, "gz");
+    ///
+    /// let (encoding, extension) = AssetEncoding::Deflate.default();
+    /// assert_eq!(encoding, AssetEncoding::Deflate);
+    /// assert_eq!(extension, "zz");
+    /// ```
+    pub fn default(self) -> (AssetEncoding, String) {
```
przydatek, line 454: nit: I like the new approach, but maybe rename `default()` to something else, say `default_extension()` or `default_config()`, as `default()` is "reserved" by `Default`-trait.

packages/ic-asset-certification/src/asset_config.rs
```diff
@@ -324,6 +403,97 @@ pub enum AssetRedirectKind {
...
+    ///
+    /// ```
+    /// use ic_asset_certification::AssetEncoding;
+    ///
+    /// let (encoding, extension) = AssetEncoding::Brotli.custom("brotli".to_string());
+    ///
+    /// assert_eq!(encoding, AssetEncoding::Brotli);
+    /// assert_eq!(extension, "brotli");
+    /// ```
+    pub fn custom(self, extension: String) -> (AssetEncoding, String) {
```
przydatek, line 479: nit: if `default()` above is renamed, then `custom()` should follow the same pattern.

packages/ic-asset-certification/src/asset_config.rs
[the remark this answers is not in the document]
przydatek: Ah, my bad, I've missed the context, sorry!

przydatek (approved):
LGTM, thanks!

packages/ic-asset-certification/src/asset_config.rs
nathanosdev: Done, thanks!

PR #345, opened 2024-07-18
fix: return certificate headers in v2 responses

r-birkner:


packages/ic-response-verification/src/verification/verify_request_response_pair.rs
```diff
@@ -272,14 +273,23 @@ fn v2_verification(
...
-        true => Ok(VerificationInfo {
-            response: Some(VerifiedResponse {
-                status_code: Some(response.status_code),
-                headers: response_headers.headers,
-                body: response.body,
-            }),
-            verification_version: 2,
-        }),
+        true => {
+            let certificate_headers: Vec<(String, String)> = vec![(
```
przydatek, line 277: nit: the variable should probably be named `certificate_header` (i.e. singular), as there is only one header.
Also, one could skip creation of the vector with one element, and just use `all_headers.push(...)`

packages/ic-response-verification/src/verification/verify_request_response_pair.rs
```diff
@@ -272,14 +273,23 @@ fn v2_verification(
...
+                CERTIFICATE_HEADER_NAME.to_string(),
+                response_headers.certificate.unwrap(),
```
przydatek, line 279: Maybe use `expect(..)` instead of `unwrap()`?
Or even better, return an error if `certificate` is `None`?

packages/ic-response-verification/src/verification/verify_request_response_pair.rs
```diff
@@ -46,11 +47,11 @@ pub fn verify_request_response_pair(
         .get("content-encoding")
         .map(|encoding| encoding.as_str());
 
-    let Some(certificate_header) = headers.get("ic-certificate") else {
+    let Some(certificate_header_str) = headers.get("ic-certificate") else {
```
przydatek, line 50: nit: since we now import `CERTIFICATE_HEADER_NAME`, maybe use `CERTIFICATE_HEADER_NAME.to_lowercase()` instead of `"ic-certificate"`?

packages/ic-http-certification-tests/tests/v2_response_verification_certification_scenarios.rs
```diff
@@ -408,15 +410,17 @@ mod tests {
             ),
         );
 
+        let mut expected_headers = expected_response
+            .headers
+            .iter()
+            .map(|(key, value)| (key.to_lowercase(), String::from(value)))
```
przydatek, line 416: nit: as we don't filter anymore, `to_lowercase()` seems obsolete here. Could we just clone() `expected_response.headers` rather than iterating through them?

przydatek (commented):
Looks good, but I left some minor questions/comments, mainly for my understanding.

packages/ic-http-certification-tests/tests/v2_response_verification_certification_scenarios.rs
r-birkner: We still need the headers to be lowercase as the "verified" response will have all the headers in lowercase. The filtering still happens during the verification.

przydatek (approved):
LGTM for .rs-code, but I suppose @nathanosdev will review .ts-code.

PR #349, opened 2024-07-24
chore: remove outdated crate categories

nathanosdev:
The release pipeline has [failed](https://github.com/dfinity/response-verification/actions/runs/10078240807/job/27862694811) because the `hash` category seems to have been removed from crates.io, although I can't find any explanation as to when or why that happened.

I took the time to review the categories of all crates just in case there's other examples, but there wasn't. So not all changes are necessary but I think the categories make more sense now.

### DIFF
--- a/packages/ic-cbor/Cargo.toml
+++ b/packages/ic-cbor/Cargo.toml
@@ -3,7 +3,7 @@ name = "ic-cbor"
 description = "CBOR decoding for Internet Computer clients"
 readme = "README.md"
 documentation = "https://docs.rs/ic-cbor"
-categories = ["api-bindings, "algorithms", "cryptography::cryptocurrencies"]
+categories = ["api-bindings", "algorithms", "cryptography::cryptocurrencies"]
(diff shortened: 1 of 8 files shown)


packages/ic-cbor/Cargo.toml
```diff
@@ -3,7 +3,7 @@ name = "ic-cbor"
 description = "CBOR decoding for Internet Computer clients"
 readme = "README.md"
 documentation = "https://docs.rs/ic-cbor"
-categories = ["api-bindings", "authentication", "cryptography"]
+categories = ["api-bindings, "algorithms", "cryptography::cryptocurrencies"]
```
przydatek, line 6: missing `"` after `api-bindings`

przydatek (approved):
LGTM, modulo a typo.

packages/ic-cbor/Cargo.toml
nathanosdev: Thanks!

PR #351, opened 2024-08-08
feat: improve http request and response types

nathanosdev:


packages/certificate-verification-js/package.json
```diff
@@ -17,7 +17,7 @@
   "files": [
     "dist"
   ],
-  "main": "dist/certificate-verification.js",
+  "main": "dist/certificate-verification.umd.cjs",
```
nathanosdev, line 20: This is an unrelated change, this was referencing the wrong file before.

packages/ic-response-verification/src/verification/body.rs
```diff
@@ -4,10 +4,10 @@ use std::io::Read;
 
 const MAX_CHUNK_SIZE_TO_DECOMPRESS: usize = 1_024;
 
-pub fn decode_body(body: &Vec<u8>, encoding: Option<&str>) -> ResponseVerificationResult<Vec<u8>> {
+pub fn decode_body(body: &[u8], encoding: Option<&str>) -> ResponseVerificationResult<Vec<u8>> {
```
nathanosdev, line 7: Unrelated change, small nit that I noticed.

packages/ic-http-certification/src/http/http_request.rs
```diff
@@ -1,39 +1,766 @@
...
-pub struct HttpRequest {
+pub struct HttpRequest<'a> {
     /// HTTP request method.
-    pub method: String,
-    /// Request URL.
-    pub url: String,
+    method: Cow<'a, str>,
+
+    /// HTTP request URL.
+    url: Cow<'a, str>,
```
przydatek, line 34: As discussed, using `Cow` for these fields seems like an overkill, that complicates the code unnecessarily.  If that's fine for you, maybe revert to `String`?

packages/ic-http-certification/src/http/http_request.rs
```diff
@@ -1,39 +1,766 @@
 use crate::{HeaderField, HttpCertificationError, HttpCertificationResult};
 use candid::{CandidType, Deserialize};
 use http::Uri;
+use std::borrow::Cow;
 
-/// A Candid-encodable representation of an HTTP request.
-/// This struct is used by canisters that implement the HTTP interface required by the HTTP Gateway Protocol.
+/// A Candid-encodable representation of an HTTP request. This struct is used by
+/// the `http_request` method of the HTTP Gateway Protocol's Candid interface.
+///
```
przydatek, line 8: maybe mention also the provided convenience helpers like `get()` or `put()`?

packages/ic-http-certification/src/http/http_request.rs
```diff
@@ -1,39 +1,766 @@
...
+    /// # Examples
+    ///
+    /// ```
+    /// use ic_http_certification::HttpRequest;
+    ///
+    /// let request = HttpRequest::patch("/").build();
+    ///
+    /// assert_eq!(request.method(), "PATCH");
+    /// ```
+    pub fn patch(url: impl Into<Cow<'a, str>>) -> HttpRequestBuilder<'a> {
```
przydatek, line 120: Not sure whether we should have the convenience helpers also for the less-used methods, as they blow-up the API considerably... Maybe `get()`, `post()`, and `put()` would be sufficient? Wdyt?

packages/ic-http-certification/src/http/http_request.rs
```diff
@@ -1,39 +1,766 @@
...
+///     .build();
+/// let update_request = HttpUpdateRequest::from(request);
+///
+/// assert_eq!(update_request.method(), "GET");
+/// assert_eq!(update_request.url(), "/");
+/// assert_eq!(update_request.headers(), &[("X-Custom-Foo".into(), "Bar".into())]);
+/// assert_eq!(update_request.body(), &[1, 2, 3]);
+/// ```
+#[derive(Clone, Debug, CandidType, Deserialize, PartialEq, Eq)]
+pub struct HttpUpdateRequest<'a> {
```
przydatek, line 633: Not sure this is really needed, maybe let's discuss offline.

przydatek (commented):
Sorry for the delay in reviewing, please see some comments below.

packages/ic-http-certification/src/http/http_request.rs
nathanosdev: While making the change, I realized that this suggestion will make the library code slightly cleaner, but the consuming code now needs to call `to_string()` for any string references.

With Cow:
```rust
let request = HttpRequest::patch("/").build();
```

versus:
```rust
let request = HttpRequest::patch("/".to_string()).build();
```

Using `Cow` for efficiency is overkill, but it also has the benefit of making the library interface more flexible and allowing consumers to pass either references, or owned values, depending on what is most convenient for them.

What do you think?

packages/ic-http-certification/src/http/http_request.rs
nathanosdev: I've personally used GET, POST, PUT, PATCH, DELETE quite a lot. So I think adding PATCH and DELETE to that list would be nice.

OPTIONS is normally handled automatically by web server frameworks. I just realized that the HEAD method is missing anyway, but that method along with TRACE and CONNECT are methods I've personally never used.

So what do you think of removing OPTIONS, TRACE and CONNECT?

packages/ic-http-certification/src/http/http_request.rs
nathanosdev: It's not really needed in the canisters, but is used in the HTTP Gateway, currently from this crate: https://github.com/dfinity/agent-rs/blob/main/ic-utils/src/interfaces/http_request.rs#L57-L66.

I was thinking to eventually remove the HTTP stuff from that `ic-utils` crate and consolidate it into the `ic-http-certification` crate, or potentially break it out into a separate `ic-http` crate if there's a need for that.

packages/ic-http-certification/src/http/http_request.rs
przydatek: Agreed, covering also other methods that are used frequently makes sense, and yeah, I'd suggest removing the ones that are rarely used.

packages/ic-http-certification/src/http/http_request.rs
przydatek: Not sure I understand: we would have `HttpRequest::patch(url: &str)`, and then one would still make a call like `HttpRequest::patch("/").build()` (while we'd do the conversion `url.to_string()` internally in `patch()`), no?  Or am I missing something?

packages/ic-http-certification/src/http/http_request.rs
przydatek: Yeah, let's chat offline 👍

packages/ic-http-certification/src/http/http_request.rs
nathanosdev: We agreed to use owned `String` values, but expose `str` reference and use `to_string()` internally.

packages/ic-http-certification/src/http/http_request.rs
nathanosdev: Chatted offline and agreed to keep the extra structure

przydatek (approved):
LGTM, thanks!

PR #352, opened 2024-08-13
fix(ic-asset-certification): handle urls that include full domain properly

nathanosdev:
Previously if an HTTP Gateway sent a URL with a domain, the `ic-asset-certification` library would fail to find the correct asset. This should never happen with any of the HTTP Gateways that we maintain in DFINITY, but making the code more resilient would be desirable anyway.

This PR fixes that bug.


przydatek (approved):
LGTM, thanks!

PR #354, opened 2024-08-13
feat(ic-asset-certification): use response header exclusions instead of inclusions for better security

nathanosdev:
Using response header exclusions as a more secure default has been encouraged by ProdSec in previous security reviews. In this case it's really the same either way because we certify every header that's added to the response, but it simplifies the code by using an empty array of exclusions that a full array on inclusions.

I did not add support for excluding response headers since it's currently not possible to return a dynamic response header anyway. If that use case comes up then I think we should revisit the idea then.


przydatek (approved):
LGTM, thanks!

PR #356, opened 2024-08-19
feat(ic-asset-ceritication): allow deleting assets from the asset router

nathanosdev:
TT-415

Adds the ability to delete certified assets from the asset router.

### DIFF
--- a/packages/ic-asset-certification/README.md
+++ b/packages/ic-asset-certification/README.md
@@ -16,6 +16,8 @@ This is implemented in the following steps:
 3. [Inserting assets into the asset router](#inserting-assets-into-the-asset-router)
 4. [Serving assets](#serving-assets)
 
+For canisters that need it, it's also possible to [delete assets](#deleting-assets).
+
 ## Preparing assets
 
 This library is unopinionated about where assets come from, so that is not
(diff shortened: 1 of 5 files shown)


packages/ic-asset-certification/src/asset_router.rs
```diff
@@ -259,9 +259,6 @@ impl<'content> AssetRouter<'content> {
     /// Certifies multiple assets and inserts them into the router, to be served
     /// later by the [serve_asset](AssetRouter::serve_asset) method.
     ///
-    /// The assets are certified using the provided
```
nathanosdev, line 262: Small left over from a previous iteration that required consumers to pass the `HttpCertificationTree` as an argument to this function.

packages/ic-asset-certification/src/asset_router.rs
```diff
@@ -315,6 +312,59 @@ impl<'content> AssetRouter<'content> {
         Ok(())
     }
 
+    /// Deletes multiple assets from the router, including any certification for those assets.
+    ///
+    /// Depending on the configuration provided to the [certify_assets](AssetRouter::certify_assets) function,
+    /// multiple responses may be generated for the same asset. To ensure that all generated responses are deleted,
+    /// this function accepts the same configuration.
+    pub fn delete_assets<'path>(
```
nathanosdev, line 320: `delete_asset` very closely replicates the logic from `certify_assets`, but deletes assets instead of certifies them.

Likewise, `delete_asset_impl` closely reflects `certify_asset_impl`.

There are other pairs too:
- `insert_static_asset` / `delete_static_asset`
- `insert_fallback_asset` / `delete_fallback_asset`
- `insert_redirect` / `delete_redirect`

Those pairs had their common code factored out into the following functions:
- `prepare_static_asset`
- `prepare_fallback_asset`
- `prepare_redirect`

packages/ic-asset-certification/src/lib.rs
```diff
@@ -1,530 +1,770 @@
...
-    value
-        .serialize(&mut serializer)
-        .expect("Failed to serialize value");
-    serializer.into_inner()
-}
-
-add_certificate_header(&mut response, &witness, &expr_path);
-```
-*/
+//!
```
nathanosdev, line 1: This file had issues with strings like `"**/*.js"`, removing the multiline doc string in favor of multiple singleline doc strings fixed the issue. I opened an issue against Rust here: https://github.com/rust-lang/rust/issues/129685

For review purposes, all changes here match exactly the changes in `README.md`

packages/ic-response-verification/src/verification/verify_request_response_pair.rs
```diff
@@ -274,7 +274,7 @@ fn v2_verification(
 
     match are_hashes_valid {
         true => {
-            let mut all_headers = response_headers.headers.clone();
+            let mut all_headers = response_headers.headers;
```
nathanosdev, line 277: Small, unrelated nit.

packages/ic-asset-certification/src/asset_router.rs
```diff
@@ -1407,6 +2241,34 @@ mod tests {
...
+        ]
+    }
+
+    #[fixture]
+    fn not_found_html_br_body() -> Vec<u8> {
+        vec![
+            27, 48, 0, 248, 45, 14, 108, 27, 88, 249, 245, 45, 213, 3, 233, 193, 146, 199, 9, 173,
+            64, 104, 10, 230, 173, 67, 124, 216, 218, 84, 12, 93, 47, 66, 139, 48, 3, 233, 78, 128,
+            105, 198, 36, 242, 83, 62, 179, 122, 129, 33, 16, 12,
+        ]
```
przydatek, line 2269: nit: please add a comment what these vectors contain and/or how they were produced.

packages/ic-asset-certification/README.md
```diff
@@ -525,3 +527,242 @@ fn cbor_encode(value: &impl Serialize) -> Vec<u8> {
...
 add_certificate_header(&mut response, &witness, &expr_path);
 ```
+
+## Deleting assets
+
+Deleting assets is similar to (certifying them)[#inserting-assets-into-the-asset-router].
+
+Depending on the configuration provided to the [certify_assets](AssetRouter::certify_assets) function,
+multiple responses may be generated for the same asset. To ensure that all generated responses are deleted,
+the [delete_assets](AssetRouter::delete_assets) function accepts the same configuration.
```
przydatek, line 537: please extend the docs by describing what happens if the configuration passed to `delete_assets` does not match the one passed previously to `certify_asset`.

packages/ic-http-certification/src/http/http_response.rs
```diff
@@ -533,3 +533,40 @@ impl<'a> From<HttpResponse<'a>> for HttpUpdateResponse<'a> {
...
+/// let b = HttpResponse::builder()
+///     .with_status_code(200)
+///     .with_headers(vec![("Content-Type".into(), "text/plain".into())])
+///     .with_body(b"Hello, World!")
+///     .with_upgrade(false)
+///     .build();
+///
+/// assert_response_eq(a, b);
+/// ```
+pub fn assert_response_eq(a: HttpResponse, b: HttpResponse) {
```
przydatek, line 561: nit: not sure this should be a part of the main package, rather than some test-util-crate 🤔

packages/ic-asset-certification/src/lib.rs
przydatek: nit: is it worthwhile to have two copies of the documentation in a package?  Wouldn't it be sufficient to have just a comprehensive `README.md` and succint per-function comments? (Maintaining both copies in sync seems quite challenging, no?)

przydatek (approved):
Nice! And very nice tests!
LGTM, thanks!

packages/ic-asset-certification/src/asset_router.rs
nathanosdev: Adding in the next commit, thanks

packages/ic-asset-certification/README.md
nathanosdev: Included, in the next commit thanks.

packages/ic-asset-certification/src/lib.rs
nathanosdev: As we discussed offline, let's keep it for this PR and review alternatives over the next days

PR #357, opened 2024-08-28
ci: switch to shared actions

nathanosdev:
Merging this will require coordination. Since the names of the jobs have changed, we will need IDX to update the names of the required jobs so that the CI can pass. Once that change is made by IDX, no other branch will be able to pass the CI pipeline until this one is merged.


.github/workflows/update-changelog.yml
```diff
@@ -1,42 +1,14 @@
-name: Update changelog
+name: generate_changelog
```
przydatek, line 1: nit: should this file be renamed too? (for consistency)

przydatek (approved):
LGTM, thanks!

.github/workflows/update-changelog.yml
nathanosdev: Done, thanks!

przydatek (approved):

PR #360, opened 2024-09-04
feat(ic-http-certification): add add_certificate_header util function

nathanosdev:
removed serde feature from ic-http-certification, condensed serde_bytes into serde feature of ic-certification

BREAKING CHANGE: removed serde feature from ic-http-certification, condensed serde_bytes into serde feature of ic-certification.

TT-426

### DIFF
--- a/examples/http-certification/assets/README.md
+++ b/examples/http-certification/assets/README.md
@@ -245,13 +245,13 @@ fn certify_all_assets() {
 
 ## Serving assets
 
-The `serve_asset` function is responsible for serving assets. It uses the `serve_asset` function from the `AssetRouter` to serve the assets. This function returns the `HttpResponse`, a witness (`HashTree`), and an expression path. The witness and expression path is used to generate the `IC-Certificate` header, which is added to the response before returning it. To add the `IC-Certificate` header to the response, the `add_certificate_header` function from the `ic-http-certification` crate is used. This function takes the data certificate, response, witness, and expression path as arguments. The witness, expression path, and the canister's certified data are encoded using CBOR and then added to the response headers.
+The `serve_asset` function is responsible for serving assets. It uses the `serve_asset` function from the `AssetRouter` to serve the assets. This function returns the `HttpResponse`, a witness (`HashTree`), and an expression path. The witness and expression path is used to generate the `IC-Certificate` header, which is added to the response before returning it. To add the `IC-Certificate` header to the response, the `add_v2_certificate_header` function from the `ic-http-certification` crate is used. This function takes the data certificate, response, witness, and expression path as arguments. The witness, expression path, and the canister's certified data are encoded using CBOR and then added to the response headers.
 
 ```rust
 fn serve_asset(req: &HttpRequest) -> HttpResponse<'static> {
     ASSET_ROUTER.with_borrow(|asset_router| {
         if let Ok((mut response, witness, expr_path)) = asset_router.serve_asset(req) {
-            add_certificate_header(
+            add_v2_certificate_header(
                 data_certificate().expect("No data certificate available"),
                 &mut response,
                 &witness,
--- a/examples/http-certification/assets/README.md
+++ b/examples/http-certification/assets/README.md
@@ -245,50 +245,23 @@ fn certify_all_assets() {
 
 ## Serving assets
 
-The `serve_asset` function is responsible for serving assets. It uses the `serve_asset` function from the `AssetRouter` to serve the assets. This function returns the `HttpResponse`, a witness (`HashTree`), and an expression path. The witness and expression path is used to generate the `IC-Certificate` header, which is added to the response before returning it.
+The `serve_asset` function is responsible for serving assets. It uses the `serve_asset` function from the `AssetRouter` to serve the assets. This function returns the `HttpResponse`, a witness (`HashTree`), and an expression path. The witness and expression path is used to generate the `IC-Certificate` header, which is added to the response before returning it. To add the `IC-Certificate` header to the response, the `add_certificate_header` function from the `ic-http-certification` crate is used. This function takes the data certificate, response, witness, and expression path as arguments. The witness, expression path, and the canister's certified data are encoded using CBOR and then added to the response headers.
 
 ```rust
 fn serve_asset(req: &HttpRequest) -> HttpResponse<'static> {
     ASSET_ROUTER.with_borrow(|asset_router| {
         if let Ok((mut response, witness, expr_path)) = asset_router.serve_asset(req) {
-            add_certificate_header(&mut response, &witness, &expr_path);
+            add_certificate_header(
+                data_certificate().expect("No data certificate available"),
+                &mut response,
+                &witness,
+                &expr_path,
+            );
 
             response
         } else {
             ic_cdk::trap("Failed to serve asset");
         }
     })
-}
-```
-
-To add the `IC-Certificate` header to the response, the `add_certificate_header` function is used. This function takes the response, witness, and expression path as arguments. The witness, expression path, and the canister's certified data are encoded using CBOR and then added to the response headers.
-
-```rust
-const IC_CERTIFICATE_HEADER: &str = "IC-Certificate";
-fn add_certificate_header(response: &mut HttpResponse, witness: &HashTree, expr_path: &[String]) {
-    let certified_data = data_certificate().expect("No data certificate available");
-    let witness = cbor_encode(witness);
-    let expr_path = cbor_encode(&expr_path);
-
-    response.add_header((
-        IC_CERTIFICATE_HEADER.to_string(),
-        format!(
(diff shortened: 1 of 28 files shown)


examples/http-certification/assets/dfx.json
```diff
@@ -6,12 +6,12 @@
     "backend": {
       "type": "custom",
       "candid": "src/backend/backend.did",
-      "wasm": "../../../target/wasm32-unknown-unknown/release/http_certification_new_assets_backend.wasm",
```
nathanosdev, line 9: `http_certification_new_assets_backend` was meant to be a temp name that I forgot to fix

examples/http-certification/assets/README.md
```diff
@@ -245,13 +245,18 @@ fn certify_all_assets() {
 
 ## Serving assets
 
-The `serve_asset` function is responsible for serving assets. It uses the `serve_asset` function from the `AssetRouter` to serve the assets. This function returns the `HttpResponse`, a witness (`HashTree`), and an expression path. The witness and expression path is used to generate the `IC-Certificate` header, which is added to the response before returning it.
+The `serve_asset` function is responsible for serving assets. It uses the `serve_asset` function from the `AssetRouter` to serve the assets. This function returns the `HttpResponse`, a witness (`HashTree`), and an expression path. The witness and expression path is used to generate the `IC-Certificate` header, which is added to the response before returning it. To add the `IC-Certificate` header to the response, the `add_certificate_header` function from the `ic-http-certification` crate is used. This function takes the data certificate, response, witness, and expression path as arguments. The witness, expression path, and the canister's certified data are encoded using CBOR and then added to the response headers.
```
nathanosdev, line 248: Updated docs to reflect the moving of `add_certificate_header` into the library

examples/http-certification/assets/src/backend/Cargo.toml
```diff
@@ -1,22 +1,14 @@
...
-name = "http_certification_new_assets_backend"
+name = "http_certification_assets_backend"
 version = "0.1.0"
 edition = "2021"
 
 [lib]
 crate-type = ["cdylib"]
 
 [dependencies]
-anyhow.workspace = true
```
nathanosdev, line 10: Some of these dependencies were already redundant, must have been some copy pasta at some point. Some dependencies are now redundant because `add_certificate_header` moved into the library

examples/http-certification/assets/src/backend/Cargo.toml
```diff
@@ -1,22 +1,14 @@
...
 candid.workspace = true
 ic-cdk.workspace = true
-ic-cdk-macros.workspace = true
-serde.workspace = true
-serde_cbor.workspace = true
-sha2.workspace = true
-ic-certification.workspace = true
-ic-http-certification = { workspace = true, features = ["serde"] }
```
nathanosdev, line 18: The `serde` feature was removed because it should always be enabled in order for `add_certificate_header` to work.

examples/http-certification/custom-assets/README.md
```diff
@@ -157,7 +157,7 @@ fn create_asset_response(
         ("cross-origin-embedder-policy".to_string(), "require-corp".to_string()),
         ("cross-origin-opener-policy".to_string(), "same-origin".to_string()),
         ("content-length".to_string(), body.len().to_string()),
-        (IC_CERTIFICATE_EXPRESSION_HEADER.to_string(), cel_expr),
+        (CERTIFICATE_EXPRESSION_HEADER_NAME.to_string(), cel_expr),
```
nathanosdev, line 160: Removing duplicate references to this const.

packages/ic-certification/Cargo.toml
```diff
@@ -32,4 +32,5 @@ serde_cbor.workspace = true
 rstest.workspace = true
 
 [features]
-default = ['serde', 'serde_bytes']
```
nathanosdev, line 35: It didn't make sense to have two separate features before, so it's combined into one now.

packages/ic-http-certification/Cargo.toml
```diff
@@ -14,17 +14,16 @@ repository.workspace = true
...
-serde = ["ic-certification/serde", "ic-certification/serde_bytes"]
-
 [dependencies]
 candid.workspace = true
 serde.workspace = true
 http.workspace = true
 urlencoding.workspace = true
 ic-representation-independent-hash.workspace = true
-ic-certification.workspace = true
+ic-certification = { workspace = true, features = ["serde"] }
```
nathanosdev, line 23: We always need this enabled for the `add_certificate_header` function

packages/ic-http-certification/src/utils/response_header.rs
```diff
@@ -0,0 +1,95 @@
+use crate::{HttpResponse, CERTIFICATE_HEADER_NAME};
+use base64::{engine::general_purpose::STANDARD as BASE64, Engine};
+use ic_certification::HashTree;
+use serde::Serialize;
+
+/// Adds the `IC-Certificate` header to a given [`HttpResponse`]. This header is used by the HTTP Gateway
+/// to verify the authenticity of query call responses.
```
przydatek, line 7: Maybe add a link to the spec (e.g. https://internetcomputer.org/docs/current/references/http-gateway-protocol-spec/#the-certificate-header), and mention that it does not perform any validity/sanity/consistency checks on the arguments?(in particular `data_certificate` and `expr_path` can have arbitrary garbage, but the function will not fail)

packages/ic-http-certification/src/utils/response_header.rs
```diff
@@ -0,0 +1,95 @@
...
+///     Retrieved using `ic_cdk::api::data_certificate`.
+/// * `response` - The [`HttpResponse`] to add the certificate header to.
+///     Created using [`HttpResponse::builder()`](crate::HttpResponse::builder).
+/// * `witness` - A pruned merkle tree revealing the relevant certification for the current response.
+///     Created using [`HttpCertificationTree::witness()`](crate::HttpCertificationTree::witness).
+/// * `expr_path` - An expression path for the current response informing the HTTP Gateway where the
+///     relevant certification is present in the merkle tree. Created using
+///     [`HttpCertificationPath::to_expr_path()`](crate::HttpCertificationPath::to_expr_path).
+///
+/// # Example
```
przydatek, line 21: I guess the example will be executed by CI, so in a way it it doubles as a unit test, (and also there are examples in this PR that use this function), but maybe it would be worthwhile to add a test explicitly, just to be sure about the test coverage?

packages/ic-http-certification/src/utils/response_header.rs
```diff
@@ -0,0 +1,95 @@
...
+    witness: &HashTree,
+    expr_path: &[String],
+) {
+    let witness = cbor_encode(witness);
+    let expr_path = cbor_encode(&expr_path);
+
+    response.add_header((
+        CERTIFICATE_HEADER_NAME.to_string(),
+        format!(
+            "certificate=:{}:, tree=:{}:, expr_path=:{}:, version=2",
```
przydatek, line 78: since this function has the certification version hard coded, maybe it would be prude to mention this at least in the documentation, or even in the function name?  (e.g. `add_certificate_v2_header`?)
What would be the plan if we want to introduce v3?  A breaking change or a new API?

przydatek (commented):
Great to see a PR with a negative line count! 🎉 
LGTM in general, but please see some questions below.

packages/ic-http-certification/src/utils/response_header.rs
nathanosdev: Good suggestion, thanks.

packages/ic-http-certification/src/utils/response_header.rs
nathanosdev: Indeed the examples are run by the CI pipeline and can serve as both documentation and unit tests. We can already be sure that the test covers the execution of the function, so I'm not sure if duplicating effort with additional unit tests would provide any value. Am I perhaps missing something?

packages/ic-http-certification/src/utils/response_header.rs
nathanosdev: That's a good point. It's not clear that this function only applies to v2.

I like the suggestion to rename the function. I went with `add_v2_certificate_header`, which felt a little nicer to say to me but I cannot explain why so I'm fine to change to `add_certificate_v2_header` if you feel strongly about it.

> What would be the plan if we want to introduce v3? A breaking change or a new API?

It's really difficult to say until we know what will actually change in v3. If it's simply the `version` part of the header then we could easily parameterize the function, but if somehow the structure of the header changes then maybe an entirely new function would be better.

packages/ic-http-certification/src/utils/response_header.rs
przydatek: Discussed offline and agreed to leave as is.

przydatek (approved):
LGTM, thanks!

PR #363, opened 2024-09-04
test: Add tests for assets with special characters

raymondk:
TT-358

### DIFF
--- a/packages/ic-response-verification-tests/src/main.rs
+++ b/packages/ic-response-verification-tests/src/main.rs
@@ -1,5 +1,6 @@
 use crate::agent::create_agent;
 use anyhow::{anyhow, Result};
+use core::panic;
 use ic_agent::export::Principal;
 use ic_agent::Agent;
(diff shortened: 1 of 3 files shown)


packages/ic-response-verification-tests/src/main.rs
```diff
@@ -23,12 +25,20 @@ fn get_current_time() -> u128 {
...
-    Ok(file)
+    );
+    match fs::read(path.clone()) {
+        Ok(file) => {
+            println!("Read file: {}", path);
+            Ok(file)
+        }
+        Err(e) => {
+            println!("Error reading file: {} from {}", e, path);
+            panic!()
```
nathanosdev, line 39: Why do you want to panic here instead of returning the error?

The return type is still `Result` even though there is no result being returned.

raymondk: That was me debugging yesterday... I'll fix it

nathanosdev (approved):

PR #366, opened 2024-09-11
Refactor response handling by AssetRouter, for extensibility

przydatek:
Refactor the containers for responses managed by `AssetRouter`, to enable extensiblity.


packages/ic-asset-certification/src/asset_router.rs
```diff
@@ -117,23 +117,39 @@ struct CertifiedAssetResponse<'a> {
...
+/// A key created from request data, to retrieve the corresponding response.
+#[derive(Debug, Eq, Hash, PartialEq)]
+pub struct RequestKey {
+    /// Path of the requested asset.
+    pub path: String,
+    /// The encoding of the assset.
+    pub encoding: Option<String>,
+}
+
+fn request_key(path: &str, encoding: Option<String>) -> RequestKey {
```
nathanosdev, line 133: It's only a style preference, but what do you think about implementing this as a constructor?

```rust
impl RequestKey {
  fn new(path: &str, encoding: Option<String>) -> Self {
    Self {
        path: path.to_string(),
        encoding,
    }
  }
}
```

packages/ic-asset-certification/src/asset_router.rs
```diff
@@ -117,23 +117,39 @@ struct CertifiedAssetResponse<'a> {
...
+    RequestKey {
+        path: path.to_string(),
+        encoding,
+    }
 }
 
 const IC_CERTIFICATE_EXPRESSION_HEADER: &str = "IC-CertificateExpression";
 
+fn encoding_str(maybe_encoding: Option<AssetEncoding>) -> Option<String> {
```
nathanosdev, line 142: Could we co-locate this with the other code for `AssetEncoding` in the `asset_config.rs` file?

nathanosdev (approved):
This is a nice improvement, thanks!

packages/ic-asset-certification/src/asset_router.rs
przydatek: SGTM, I'll do that in the next PR that builds on top of this (so that I don't have to reconcile merge conflicts), ok?

packages/ic-asset-certification/src/asset_router.rs
przydatek: Same here, I'll do this change in the next PR.

przydatek (commented):
Thanks, I'll merge as-is, and include the suggested changes to the next PR.

PR #369, opened 2024-09-17
feat(ic-asset-certification): integrate add_certificate_header into the library so consumers do not need to call it manually

nathanosdev:
TT-436

### DIFF
--- a/packages/ic-asset-certification/src/asset_router.rs
+++ b/packages/ic-asset-certification/src/asset_router.rs
@@ -2,11 +2,10 @@ use crate::{
     Asset, AssetCertificationError, AssetCertificationResult, AssetConfig, AssetEncoding,
     AssetFallbackConfig, AssetRedirectKind, NormalizedAssetConfig,
 };
-use ic_certification::HashTree;
 use ic_http_certification::{
-    DefaultCelBuilder, DefaultResponseCertification, Hash, HttpCertification,
-    HttpCertificationPath, HttpCertificationTree, HttpCertificationTreeEntry, HttpRequest,
-    HttpResponse, CERTIFICATE_EXPRESSION_HEADER_NAME,
+    utils::add_v2_certificate_header, DefaultCelBuilder, DefaultResponseCertification, Hash,
+    HttpCertification, HttpCertificationPath, HttpCertificationTree, HttpCertificationTreeEntry,
+    HttpRequest, HttpResponse, CERTIFICATE_EXPRESSION_HEADER_NAME,
 };
 use std::{borrow::Cow, cell::RefCell, collections::HashMap, rc::Rc};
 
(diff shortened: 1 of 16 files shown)


packages/ic-asset-certification/src/asset_router.rs
```diff
@@ -840,18 +822,105 @@ impl Default for AssetRouter<'_> {
...
+        cel::DefaultFullCelExpressionBuilder, HeaderField, CERTIFICATE_HEADER_NAME,
+    };
+    use ic_response_verification::CertificateHeader;
     use rstest::*;
     use std::vec;
 
+    // A certificate taken from a real response on mainnet. It doesn't matter what it contains,
+    // as long as it's a valid certificate. If we ever decide to run response verification in these
+    // tests then the content of the certificate will matter.
+    const DATA_CERTIFICATE: &[u8] = &[
```
przydatek, line 836: nit: maybe add these bytes as hex/base64/cbor string?  It would be shorter and and easier to handle.

packages/ic-asset-certification/src/asset_router.rs
```diff
@@ -180,92 +181,29 @@ impl<'content> AssetRouter<'content> {
...
-                let witness = self.tree.borrow().witness(tree_entry, &req_path)?;
+            }) => {
+                let witness = self.tree.borrow().witness(&tree_entry, &request_url)?;
                 let expr_path = tree_entry.path.to_expr_path();
 
-                return Ok((response.clone(), witness, expr_path));
-            }
-
-            scope.pop();
+                add_certificate_header(data_certificate, &mut response, &witness, &expr_path);
```
przydatek, line 201: If we use hard-coded v2-certification in the `AssetRouter` (via `add_certificate_header`), how will clients transition to v3?  Maybe certification version should be a property of `AssetRouter` set in the constructor?  🤔

packages/ic-asset-certification/src/asset_router.rs
```diff
@@ -180,92 +181,29 @@ impl<'content> AssetRouter<'content> {
     /// given [HttpRequest](ic_http_certification::HttpRequest).
     pub fn serve_asset(
         &self,
+        data_certificate: &[u8],
```
przydatek, line 184: Please document the added argument, maybe with a reference to the spec about the format, and mentioning `ic_cdk::api::data_certificate()` as a usual source for the value (as you've mentioned in the code example above).

przydatek (commented):
I suppose the final shape of this PR depends on #360, but please see some initial comments/questions.

packages/ic-asset-certification/src/asset_router.rs
nathanosdev: Good point, thanks.

packages/ic-asset-certification/src/asset_router.rs
przydatek: Let's discuss that next week, and maybe address in a separate PR.

przydatek (approved):
LGTM, thanks!

PR #373, opened 2024-09-19
feat(@dfinity/response-verification): unify request and response types with the corresponding HTTP Gateway types

nathanosdev:
The `request` and `response` types used by the `@dfinity/response-verification` JS package (JS wrapper around `ic-response-verification` WASM) was almost, but not exactly compatible with the corresponding request and response types use by the HTTP Gateway, resulting in the need to map between the two types. This PR updates those types so that they're not compatible.

TT-438


przydatek (approved):
LGTM, thanks!
(the PR-description says "This PR updates those types so that they're not compatible.", I guess what is meant is "... so they're compatible", right?)

nathanosdev:
> LGTM, thanks! (the PR-description says "This PR updates those types so that they're not compatible.", I guess what is meant is "... so they're compatible", right?)

Yes, I meant to write "now compatible" instead of "not compatible". Sorry for the confusion.

PR #382, opened 2024-10-08
docs: add 'upgrade to update call' example

nathanosdev:


examples/http-certification/upgrade-to-update-call/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+Since browsers are unable to directly interact with the IC network, the HTTP Gateway acts as a bridge between the two. The HTTP Gateway forwards requests from clients to canisters and forwards responses from canisters back to clients. Before returning responses from canister back to clients, the HTTP Gateway verifies the certification of the response to ensure that they have not been tampered with.
+
+Upgrading query calls to upgrade calls allows for the certification of any kind of dynamic response by leveraging ICP's consensus protocol without having to statically certify the response ahead of time. This is the simplest way to add _secure_ HTTP support to a canister.
+
+A similairly simple yet more performant, but _insecure_ approach is to skip certification entirely. This is not recommended unless you are absolutely sure that certification really does not make sense for your canister. Check the ["Skipping certification for HTTP responses"](https://internetcomputer.org/docs/current/developer-docs/web-apps/http-compatible-canisters/skipping-certification-for-http-responses) guide for more details on how to do that.
```
przydatek, line 11: The link to "Skipping certification (...)" seems to be broken.

examples/http-certification/upgrade-to-update-call/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+This is not a beginner's canister development guide. Many foundational concepts that a relatively experienced canister developer should already know will be omitted. Concepts specific to upgrading to an update call will be called out here and can help to understand the [full code example](https://github.com/dfinity/response-verification/tree/main/examples/http-certification/upgrade-to-update-call).
```
przydatek, line 13: nit: I'd use "fundamental" rather than "foundational" (but I'm not a native speaker :wink:)

examples/http-certification/upgrade-to-update-call/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+Upon receiving a response from the canister with the `upgrade` field set to `opt true`, the HTTP Gateway will repeat the original request as an update call to the `http_request_update` method of the canister. The canister can then respond to the update call with any dynamic response and leverage the ICP consensus protocol for security. The certification resulting from putting this response through consensus will be verified by the HTTP Gateway to ensure it has not been tampered with.
+
+## Rust
+
+This example project features both Rust and Motoko code. If you rather follow the Motoko version, you can skip this section and go straight to the [section covering Motoko](#motoko).
+
+The Rust code is split into two functions: `http_request` and `http_request_update`. The `http_request` function is the entry point for the query call from the HTTP Gateway. It returns an `HttpResponse` with the `upgrade` field set to `true`. The `http_request_update` function is the entry point for the update call from the HTTP Gateway. It returns an `HttpUpdateResponse` with a custom status code and body.
```
przydatek, line 25: nit: technically, `upgrade`-field is set to `Some(true)`, by calling `with_upgrade(true)`-method, which seems more inline with `opt true` used in the section "How it works" above.  

Also, maybe it would be worthwhile to mention that `upgrade` set to `Some(false)` has the same semantics as `None`?

przydatek (approved):
LGTM, thanks!

examples/http-certification/upgrade-to-update-call/README.md
nathanosdev: That's expected for now, because the page doesn't exist. I'll need to make another PR on the docs repo to create that page.

examples/http-certification/upgrade-to-update-call/README.md
nathanosdev: Interesting, I didn't think about this when I wrote it. But I asked ChatGPT for advice and it gave me the following:

The distinction between "foundational concepts" and "fundamental concepts" is subtle, but they can be understood in different contexts:

1. **Foundational Concepts**:
   - **Role**: These are concepts that serve as the base or groundwork for a system of thought, discipline, or framework. They help build more complex ideas and are often specific to a certain field or structure. 
   - **Scope**: These concepts may not be universally applicable across all domains but are essential for understanding a particular area or system.
   - **Example**: In mathematics, the concept of "number" or "set" is foundational because it's critical to build the rest of the discipline.

2. **Fundamental Concepts**:
   - **Role**: These refer to basic, core ideas that are essential and universal to understanding broader or multiple systems of thought. They are more intrinsic and represent essential truths or properties that underlie a field.
   - **Scope**: Fundamental concepts often have broad applicability and are seen as essential truths or principles across different disciplines.
   - **Example**: In physics, the concept of "energy" or "force" is fundamental because it applies universally across many different branches of physics.

### Key Difference:
- **Foundational concepts** are more about what forms the basis for further learning or development in a particular area.
- **Fundamental concepts** are about what is universally necessary or basic across multiple systems or domains.

In practice, people may use these terms interchangeably, but the emphasis can shift based on whether they are talking about a specific system (foundational) or universal ideas (fundamental).

examples/http-certification/upgrade-to-update-call/README.md
nathanosdev: Very good point, thanks!

examples/http-certification/upgrade-to-update-call/README.md
nathanosdev: I updated it to fundamental anyway, I think it sounds nicer to say

PR #391, opened 2024-11-05
fix(ic-asset-certification): remove dependency on regex to reduce WASM size

przydatek:
Using `regex`-crate increases the WASM-size of canisters by over 100 kB. This PR replaces `regex` with a custom function for "manual" parsing of `Range`-header.  For example the canister `http_gateway_canister_custom_assets` with `regex` had size 466'631 bytes, while with the custom function only 344'593 bytes.

[TT-440](https://dfinity.atlassian.net/browse/TT-440)

[TT-440]: https://dfinity.atlassian.net/browse/TT-440?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Cargo.toml
```diff
@@ -62,9 +62,9 @@ http = "0.2"
 flate2 = "1.0"
 sha2 = "0.10"
 urlencoding = "2.1"
-regex = "1"
 rstest = "0.18"
 rstest_reuse = "0.2"
+text_io = "0.1"
```
nathanosdev, line 67: This library was a great find! It seems to be unmaintained though, and hasn't had any meaningful updates in a couple of years. I spotted this comment from the author on one of the recent issues too: https://github.com/oli-obk/rust-si/issues/48#issuecomment-1979636446.

> Unfortunately I have not maintained this crate in years, I don't even remember how it works in detail, so unless someone comes along and reproduces your issue and fixes it, I don't think it'll get fixed in the next few years

The library is very simple though, and we don't even need all of the code. What do you think of extracting the code we need so that we can maintain it oureslves?

packages/ic-asset-certification/src/asset_router.rs
```diff
@@ -133,19 +133,20 @@ fn encoding_str(maybe_encoding: Option<AssetEncoding>) -> Option<String> {
 }
 
 fn parse_range_header_str(str_value: &str) -> Result<RangeRequestValues, String> {
-    // expected format: `bytes <range-begin>-[<range-end>]`
-    let re = Regex::new(r"bytes=(\d+)-(\d*)").expect("internal: wrong RE");
-    let Some(caps) = re.captures(str_value) else {
-        return Err("malformed Range header".to_string());
+    // expected format: `bytes=<range-begin>-[<range-end>]`
+    let range_begin: usize = try_read!("bytes={}-", str_value.bytes())
```
nathanosdev, line 137: Does it work to use `try_scan!("bytes={}-{}", str_value.bytes())`?

przydatek: The problem is the optional `range-end` -- if it is missing, we should not complain.

Cargo.toml
nathanosdev: That's true, their code is very generic and does a lot of stuff we don't actually need so maintaining that code would be overkill. 

But thinking about it further, I'm not sure if we even need an external library, what do you think of using something like this instead?

```rust
pub fn extract_range_header_values(range_header: &str) -> Result<(u64, Option<u64>), String> {
    let range_header = range_header.trim();
    if !range_header.starts_with("bytes=") {
        return Err("Invalid range header".to_string());
    }

    let range_header = range_header.trim_start_matches("bytes=");
    let range_header_parts = range_header.split('-').collect::<Vec<_>>();
    if range_header_parts.len() == 0 || range_header_parts.len() > 2 {
        return Err("Invalid range header".to_string());
    }

    let start = range_header_parts[0]
        .parse::<u64>()
        .map_err(|_| "Invalid range header".to_string())?;

    let end = match range_header_parts[1] {
        "" => None,
        _ => Some(
            range_header_parts[1]
                .parse::<u64>()
                .map_err(|_| "Invalid range header".to_string())?,
        ),
    };

    Ok((start, end))
}

#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn one_value() {
        let result = extract_range_header_values("bytes=500-").unwrap();
        assert_eq!(result, (500, None));
    }

    #[test]
    fn two_values() {
        let result = extract_range_header_values("bytes=0-499").unwrap();
        assert_eq!(result, (0, Some(499)));
    }

    #[test]
    fn invalid_range_header() {
        let result = extract_range_header_values("bytes=500-0-499");
        assert_eq!(result, Err("Invalid range header".to_string()));
    }
}
```

nathanosdev (approved):

PR #393, opened 2024-11-18
feat(ic-asset-certification): add function to delete all assets

nathanosdev:


.github/workflows/check-commit-messages.yml
```diff
@@ -26,4 +26,4 @@ jobs:
           pip install -U Commitizen
 
       - name: Check Commit Messages
-        run: cz check --rev-range c8ecbc19b8c4a482e55907d37554d66f2f2f9a8f..HEAD
+        run: cz check --rev-range ac8f90b7c45e13e240d6c01a43d191b99f1aec4a..HEAD
```
nathanosdev, line 29: I accidentally merged the "update changelog" PR with the wrong commit message so now we have to check commit messages from a later commit starting point

packages/ic-asset-certification/README.md
```diff
@@ -750,6 +750,16 @@ use ic_cdk::api::set_certified_data;
 set_certified_data(&asset_router.root_hash());
 ```
 
+It's also possible to delete all assets and their certification in one go:
+
+```rust
+# use ic_asset_certification::{Asset, AssetConfig, AssetFallbackConfig, AssetRouter, AssetRedirectKind, AssetEncoding};
```
przydatek, line 756: nit: why do we need to use all these objects when the code is using AssetRouter only?

packages/ic-asset-certification/src/asset_router.rs
```diff
@@ -269,6 +269,9 @@ impl<'content> AssetRouter<'content> {
     ///
     /// If no configuration matches an individual asset, the asset will be
     /// served and certified as-is, without headers.
+    ///
+    /// After performing this operation, the canister's certified variable will need to be updated
```
przydatek, line 273: nit: I'd rephrase to something more direct, e.g. "After this operation one must set canister's certified data (`ic_cdk::set_certified data()`) to the new [root hash](AssetRouter::root_hash) of the tree." 
(also for the other functions below)

packages/ic-asset-certification/src/asset_router.rs
```diff
@@ -2689,6 +2705,25 @@ mod tests {
         ));
     }
 
+    #[rstest]
+    fn test_delete_all_assets() {
+        let mut asset_router = asset_router();
+        asset_router.delete_all_assets();
```
przydatek, line 2711: nit: maybe before deletion add a check that `/index.html` is present?

przydatek (dismissed):
LGTM, thanks!

packages/ic-asset-certification/README.md
nathanosdev: Copy pasta error, thanks!

przydatek (approved):
LGTM, thanks!

PR #397, opened 2024-11-28
feat(ic-http-certification): add request method enum

nathanosdev:
I took a very similar approach here as I did for response status code. Thankfully there's a lot less sweeping changes in this case though.

Additionally I noticed that I missed some docs that should've been updated with the status code change, so I fixed them too.

### DIFF
--- a/examples/http-certification/json-api/README.md
+++ b/examples/http-certification/json-api/README.md
@@ -142,7 +142,7 @@ Some headers from this project have not been included:
 To facilitate the consistent usage of these headers, there is a reusable `create_response` function used when creating responses:
 
 ```rust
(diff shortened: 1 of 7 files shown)


examples/http-certification/json-api/README.md
```diff
@@ -270,25 +270,32 @@ fn certify_list_todos_response() {
...
-                .with_method(method)
-                .with_url(TODOS_PATH)
-                .build();
-
-            let body = ErrorResponse::not_allowed().encode();
-            let mut response = create_response(405, body);
-
-            certify_response(request, &mut response, &TODOS_TREE_PATH);
-        });
+    [
```
nathanosdev, line 279: It's only the request methods that changed here, the rest is formatting.

przydatek (approved):
LGTM, thanks!

PR #398, opened 2024-11-28
feat(ic-http-certification): add http response convenience methods

nathanosdev:
Most changes are just using the new convenience methods, I've commented the main file were the meaningful changes are.

### DIFF
--- a/packages/ic-http-certification/src/http/http_response.rs
+++ b/packages/ic-http-certification/src/http/http_response.rs
@@ -4,6 +4,7 @@ use candid::{
     CandidType, Deserialize,
 };
 pub use http::StatusCode;
+use serde::Deserializer;
 use std::{borrow::Cow, fmt::Debug};
 
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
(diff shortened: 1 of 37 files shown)


packages/ic-http-certification/src/http/http_response.rs
```diff
@@ -62,6 +62,34 @@ impl From<StatusCode> for StatusCodeWrapper {
...
+/// - [NO_CONTENT](HttpResponse::no_content)
+/// - [MOVED_PERMANENTLY](HttpResponse::moved_permanently)
+/// - [TEMPORARY_REDIRECT](HttpResponse::temporary_redirect)
+/// - [BAD_REQUEST](HttpResponse::bad_request)
+/// - [UNAUTHORIZED](HttpResponse::unauthorized)
+/// - [FORBIDDEN](HttpResponse::forbidden)
+/// - [NOT_FOUND](HttpResponse::not_found)
+/// - [METHOD_NOT_ALLOWED](HttpResponse::method_not_allowed)
+/// - [TOO_MANY_REQUESTS](HttpResponse::too_many_requests)
+/// - [INTERNAL_SERVER_ERROR](HttpResponse::internal_server_error)
```
przydatek, line 82: Not sure whether this is not making things "too convenient", i.e. whether it is worthwhile to add so many convenience methods... 🤔  How about having at most 5 or so?  The builder-pattern is pretty easy, so it can be used directly for the less common reponse codes, while having many convenience helpers requires a substantial amount of code to maintain.

przydatek: As decided in an offline discussion we'll keep all the convenience methods for now, as apparently most of them are useful quite frequently.

przydatek (approved):
LGTM, thanks!

przydatek (dismissed):

przydatek (approved):

PR #402, opened 2024-12-05
feat(ic-asset-certification): add ability to delete assets by path

nathanosdev:


packages/ic-asset-certification/README.md
```diff
@@ -530,7 +538,7 @@ If the configuration excludes a file that was certified, it will not be deleted.
 provided to `certify_assets` includes the Brotli and Gzip encodings, but the configuration provided to `delete_assets`
 only includes Brotli, then the Gzip file will not be deleted.
 
-Using the same base example used to demonstrate certifying assets:
+Assuming the same base example used above to demonstrate certifying assets:
```
przydatek, line 541: How about adding just a reference/link to the example above instead of copying here (and one more time below)?  The snippet for deletion is in a separated box anyway, so this is not a complete example for deletion, no?

packages/ic-asset-certification/README.md
```diff
@@ -746,18 +750,191 @@ asset_router
...
+they will not be deleted.
+
+If multiple encodings exist for a path, all encodings will be deleted.
+
+Fallbacks are also not deleted, to delete them, use the
+[delete_fallback_assets_by_path](AssetRouter::delete_fallback_assets_by_path) function.
+
+Assuming the same base example used above to demonstrate certifying assets:
+
+```rust
```
przydatek, line 776: (this is the other copy of the example setup)

packages/ic-asset-certification/src/lib.rs
```diff
@@ -520,7 +527,7 @@
 //! provided to `certify_assets` includes the Brotli and Gzip encodings, but the configuration provided to `delete_assets`
 //! only includes Brotli, then the Gzip file will not be deleted.
 //!
-//! Using the same base example used to demonstrate certifying assets:
+//! Assuming the same base example used above to demonstrate certifying assets:
```
przydatek, line 530: (analogously, maybe avoid the setup-copies also in this file)

packages/ic-http-certification/src/tree/certification_tree.rs
```diff
@@ -205,7 +214,228 @@ mod tests {
                 "<$>"
             ]),
             SubtreeLookupResult::Found(_)
-        ));
+        );
+    }
+
+    #[rstest]
+    fn test_delete_by_path() {
```
przydatek, line 221: nice test! 🎉

przydatek (approved):
LGTM, thanks!

packages/ic-asset-certification/README.md
nathanosdev: I'm unsure about this, because it will mean that readers have to jump up and down the docs more and it also means that when changing the original example, we have to remember that it's also being referenced in an another, unrelated section. If we have independent examples, they can change independently, if necessary.

PR #403, opened 2024-12-09
fix(ic-http-certification): fix handling of uncertified query parameters

nathanosdev:


packages/ic-http-certification/src/hash/request_hash.rs
```diff
@@ -77,10 +80,12 @@ fn get_filtered_query(query: &str, request_certification: &DefaultRequestCertifi
                 })
                 .unwrap_or(false)
         })
-        .collect::<Vec<&str>>()
-        .join("&");
```
nathanosdev, line 81: When the vector was empty, this would return an empty string. We should actually return `None` here so that this part works as expected:

https://github.com/dfinity/response-verification/pull/403/files#diff-d7eea369c9615e923e6f32afc2e37bb4aabfa6475f401f91e2f76fab6e81e609L22-L24

przydatek (approved):
LGTM, thanks!

PR #408, opened 2024-12-18
ci: fix canister declaration generation on release pipeline

nathanosdev:


.github/workflows/check-commit-messages.yml
```diff
@@ -26,4 +26,4 @@ jobs:
           pip install -U Commitizen
 
       - name: Check Commit Messages
-        run: cz check --rev-range ac8f90b7c45e13e240d6c01a43d191b99f1aec4a..HEAD
+        run: cz check --rev-range 4b3de72e9b2a1ea5b700851c25db6f42cb55e8fa..HEAD
```
nathanosdev, line 29: We need to exclude the recent commit that does not match the spec

.github/workflows/create-release-pr.yml
```diff
@@ -48,7 +48,7 @@ jobs:
       - name: Create Pull Request
         uses: peter-evans/create-pull-request@v4
         with:
-          title: 'Release ${{ steps.cz.outputs.version }}'
+          title: 'chore: release ${{ steps.cz.outputs.version }}'
```
nathanosdev, line 51: Make sure the release PR title matches commitizen so we don't accidentally create commit messages that don't match the spec

przydatek (approved):
LGTM, thanks!

PR #417, opened 2024-12-20
fix(ic-asset-certification): include range header in streaming request certification

nathanosdev:


packages/ic-asset-certification/src/asset_router.rs
```diff
@@ -908,7 +908,10 @@ impl<'content> AssetRouter<'content> {
                 http::header::CONTENT_RANGE.to_string(),
                 format!("bytes {range_begin}-{range_end}/{total_length}"),
             ));
-            request_headers.push(http::header::RANGE.to_string());
+            request_headers.push((
+                http::header::RANGE.to_string(),
+                format!("bytes={range_begin}-"),
```
nathanosdev, line 913: Include the value of the header, not just the name

packages/ic-asset-certification/src/asset_router.rs
```diff
@@ -925,7 +928,7 @@ impl<'content> AssetRouter<'content> {
         status_code: StatusCode,
         body: Cow<'content, [u8]>,
         additional_response_headers: Vec<(String, String)>,
-        certified_request_headers: Vec<String>,
+        certified_request_headers: Vec<(String, String)>,
```
nathanosdev, line 931: Expect a vector of header names and values, instead of just names

packages/ic-asset-certification/src/asset_router.rs
```diff
@@ -934,7 +937,7 @@ impl<'content> AssetRouter<'content> {
             .with_request_headers(
                 certified_request_headers
                     .iter()
-                    .map(|s| s.as_str())
+                    .map(|(s, _)| s.as_str())
```
nathanosdev, line 940: Make sure to take just the header name for the CEL expr

packages/ic-asset-certification/src/asset_router.rs
```diff
@@ -944,7 +947,9 @@ impl<'content> AssetRouter<'content> {
         let cel_expr_str = cel_expr.to_string();
         headers.push((CERTIFICATE_EXPRESSION_HEADER_NAME.to_string(), cel_expr_str));
 
-        let request = HttpRequest::get(url).build();
+        let request = HttpRequest::get(url)
+            .with_headers(certified_request_headers.clone())
```
nathanosdev, line 951: and then include the full header name/value pair in the request for calculating the request hash

r-birkner (approved):

PR #420, opened 2025-01-22
fix(ic-asset-certification): fix range response certification

nathanosdev:
add additional error types to ic-response-verification crate

### DIFF
--- a/packages/ic-asset-certification/src/asset_router.rs
+++ b/packages/ic-asset-certification/src/asset_router.rs
@@ -908,10 +908,15 @@ impl<'content> AssetRouter<'content> {
                 http::header::CONTENT_RANGE.to_string(),
                 format!("bytes {range_begin}-{range_end}/{total_length}"),
             ));
-            request_headers.push((
-                http::header::RANGE.to_string(),
-                format!("bytes={range_begin}-"),
-            ));
+
+            // The `Range` request header will not be sent with the first request,
+            // so we don't include it in certification for the first chunk.
+            if range_begin != 0 {
+                request_headers.push((
+                    http::header::RANGE.to_string(),
+                    format!("bytes={range_begin}-"),
+                ));
+            }
         };
 
         Self::prepare_response_and_certification(
@@ -1260,7 +1265,7 @@ mod tests {
         let request = HttpRequest::get(&req_url).build();
         let mut expected_response = build_206_response(
             asset_body[0..ASSET_CHUNK_SIZE].to_vec(),
-            asset_range_chunk_cel_expr(),
+            asset_cel_expr(),
             vec![
                 (
                     "cache-control".to_string(),
@@ -1460,7 +1465,7 @@ mod tests {
             .build();
         let mut expected_response = build_206_response(
             asset_body[0..ASSET_CHUNK_SIZE].to_vec(),
-            encoded_range_chunk_asset_cel_expr(),
+            asset_cel_expr(),
             vec![
                 (
                     "cache-control".to_string(),
@@ -3085,7 +3090,7 @@ mod tests {
                 .get(format!("/{}", TWO_CHUNKS_ASSET_NAME), None, Some(0));
         let expected_first_chunk_response = build_206_response(
             first_chunk_body.to_vec(),
-            asset_range_chunk_cel_expr(),
+            asset_cel_expr(),
             vec![
                 (
                     "cache-control".to_string(),
@@ -3111,7 +3116,7 @@ mod tests {
         );
         let expected_first_chunk_gzip_response = build_206_response(
             first_chunk_gzip_body.to_vec(),
-            asset_range_chunk_cel_expr(),
+            encoded_asset_cel_expr(),
             vec![
                 (
                     "cache-control".to_string(),
@@ -3251,7 +3256,7 @@ mod tests {
         let first_chunk_body = &full_body[0..ASSET_CHUNK_SIZE];
         let expected_first_chunk_response = build_206_response(
             first_chunk_body.to_vec(),
-            asset_range_chunk_cel_expr(),
+            asset_cel_expr(),
             vec![
                 (
                     "cache-control".to_string(),
--- a/packages/ic-asset-certification/tests/common/mod.rs
+++ b/packages/ic-asset-certification/tests/common/mod.rs
@@ -1,8 +1,11 @@
 use ic_asset_certification::ASSET_CHUNK_SIZE;
 use ic_response_verification_test_utils::hash;
-use rand_chacha::{rand_core::{RngCore, SeedableRng}, ChaCha20Rng};
+use rand_chacha::{
+    rand_core::{RngCore, SeedableRng},
+    ChaCha20Rng,
+};
 
-pub fn asset_chunk<'a>(asset_body: &'a [u8], chunk_number: usize) -> &'a [u8] {
+pub fn asset_chunk(asset_body: &[u8], chunk_number: usize) -> &[u8] {
     let start = chunk_number * ASSET_CHUNK_SIZE;
     let end = start + ASSET_CHUNK_SIZE;
     &asset_body[start..end.min(asset_body.len())]
--- a/packages/ic-asset-certification/tests/common/mod.rs
+++ b/packages/ic-asset-certification/tests/common/mod.rs
@@ -1,0 +1,55 @@
+use ic_asset_certification::ASSET_CHUNK_SIZE;
+use ic_response_verification_test_utils::hash;
+use rand_chacha::{rand_core::{RngCore, SeedableRng}, ChaCha20Rng};
+
+pub fn asset_chunk<'a>(asset_body: &'a [u8], chunk_number: usize) -> &'a [u8] {
+    let start = chunk_number * ASSET_CHUNK_SIZE;
+    let end = start + ASSET_CHUNK_SIZE;
+    &asset_body[start..end.min(asset_body.len())]
+}
+
+pub fn asset_body(asset_name: &str, asset_size: usize) -> Vec<u8> {
+    let mut rng = ChaCha20Rng::from_seed(hash(asset_name));
+    let mut body = vec![0u8; asset_size];
+    rng.fill_bytes(&mut body);
+
+    body
+}
+
+#[macro_export]
+macro_rules! assert_contains {
+    ($vec:expr, $elems:expr) => {
+        for elem in $elems {
+            assert!(
+                $vec.contains(&elem),
+                "assertion failed: Expected vector {:?} to contain element {:?}",
+                $vec,
+                elem
+            );
+        }
+    };
+}
+
+#[macro_export]
+macro_rules! assert_response_eq {
+    ($actual:expr, $expected:expr) => {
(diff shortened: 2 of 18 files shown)


packages/ic-asset-certification/src/asset_router.rs
```diff
@@ -908,10 +908,15 @@ impl<'content> AssetRouter<'content> {
                 http::header::CONTENT_RANGE.to_string(),
                 format!("bytes {range_begin}-{range_end}/{total_length}"),
             ));
-            request_headers.push((
-                http::header::RANGE.to_string(),
-                format!("bytes={range_begin}-"),
-            ));
+
```
nathanosdev, line 911: This is the main change. We don't want to add the range header to certification for the first request because the gateway doesn't send it, otherwise this request/response pair will fail response verification.

packages/ic-asset-certification/tests/common/mod.rs
```diff
@@ -0,0 +1,55 @@
+use ic_asset_certification::ASSET_CHUNK_SIZE;
```
nathanosdev, line 1: I've thought for a while that I'd like to split the tests in the asset router. Any tests that work with the public API of the crate can be moved out as "integration" tests and allow us to split them into different files while keeping a common file to share between tests. In this PR I've only added the new tests in this way, but I'd like to follow up to further refactor the other tests in this way.

The new tests focus on testing the bug fix, that the first chunk and second chunk are returned as expected and pass response verification.

Response verification was missing from the tests in this crate up until now.

packages/ic-asset-certification/tests/common/mod.rs
```diff
@@ -0,0 +1,55 @@
...
+pub fn asset_body(asset_name: &str, asset_size: usize) -> Vec<u8> {
+    let mut rng = ChaCha20Rng::from_seed(hash(asset_name));
+    let mut body = vec![0u8; asset_size];
+    rng.fill_bytes(&mut body);
+
+    body
+}
+
+#[macro_export]
+macro_rules! assert_contains {
```
nathanosdev, line 20: Writing macros for this is a bit of an experiment, but it seems to work well. It makes the tests cleaner in my opinion and gives reasonable error messages for the correct line number (meaning the line number of where the macro is called from, rather than the line number of the implementation) when it fails.

packages/ic-asset-certification/tests/large_assets.rs
```diff
@@ -0,0 +1,193 @@
...
+    let canister_id = create_canister_id("rdmx6-jaaaa-aaaaa-aaadq-cai");
+    let req_url = format!("/{}", ASSET_ONE_NAME);
+
+    let mut asset_router = AssetRouter::default();
+    let assets = [Asset::new(ASSET_ONE_NAME, asset_one_body)];
+    let asset_configs = [asset_config(ASSET_ONE_NAME.to_string(), vec![])];
+    asset_router.certify_assets(assets, asset_configs).unwrap();
+
+    let certified_data = asset_router.root_hash();
+    let CertificateData {
```
nathanosdev, line 37: This is one of the main differences between this test and the existing tests. We create a full certificate with a valid root key and use this to create a fully verifiable `IC-Certificate` header.

packages/ic-asset-certification/tests/large_assets.rs
```diff
@@ -0,0 +1,193 @@
...
+        .with_body(asset_one_chunk_one)
+        .build();
+
+    let chunk_one_req = HttpRequest::get(&req_url).build();
+    let chunk_one_res = asset_router
+        .serve_asset(&cbor_encoded_certificate, &chunk_one_req)
+        .unwrap();
+    assert_response_eq!(chunk_one_res, expected_chunk_one_res);
+
+    let chunk_one_verification = verify_request_response_pair(
```
nathanosdev, line 68: And then of course running response verification is the other main difference.

packages/ic-asset-certification/tests/large_assets.rs
```diff
@@ -0,0 +1,193 @@
...
+        .with_response_certification(DefaultResponseCertification::response_header_exclusions(
+            vec![],
+        ))
+        .build()
+        .to_string()
+}
+
+#[fixture]
+fn asset_one_body() -> &'static Vec<u8> {
+    static ASSET_ONE_BODY: OnceCell<Vec<u8>> = OnceCell::new();
```
nathanosdev, line 158: Another experiment here, to try and improve the performance of tests with large assets. In this case it only makes a minor difference, but it should make a bigger difference for tests with larger files, especially when we reuse the same assets across multiple tests.

packages/ic-http-certification-tests/tests/v1_response_verification.rs
```diff
@@ -63,13 +64,13 @@ mod tests {
         )
         .unwrap();
 
-        assert!(matches!(
+        assert_matches!(
```
nathanosdev, line 67: There's also some unrelated, minor refactoring.

packages/ic-http-certification/src/http/http_request.rs
```diff
@@ -287,6 +287,26 @@ impl<'a> HttpRequest<'a> {
...
+    /// let mut request = HttpRequest::get("/")
+    ///     .with_headers(vec![("Content-Type".into(), "text/plain".into())])
+    ///     .build();
+    ///
+    /// request.headers_mut().push(("Content-Length".into(), "13".into()));
+    ///
+    /// assert_eq!(request.headers(), &[("Content-Type".into(), "text/plain".into()), ("Content-Length".into(), "13".into())]);
+    /// ```
+    #[inline]
+    pub fn headers_mut(&mut self) -> &mut Vec<HeaderField> {
```
nathanosdev, line 306: This was needed on the http gateway to modify the headers of the request and insert the range header. We already have the same method for the response so I think it was an oversight to omit that anyway.

packages/ic-response-verification/src/error.rs
```diff
@@ -49,11 +49,15 @@ pub enum ResponseVerificationError {
 
     /// The tree has different root hash from the expected value in the certified variables
     #[error("Invalid tree root hash")]
-    InvalidTree,
-
-    /// The CEL expression path is invalid
-    #[error("Invalid expression path")]
-    InvalidExpressionPath,
+    InvalidTreeRootHash,
```
nathanosdev, line 52: The other change I've made in this PR is add more granular errors. This really helped me to debug this issue.

przydatek (commented):
LGTM in general, but please see some questions below.
(Thanks a lot for the annotations to the changes, very helpful indeed!)

packages/ic-asset-certification/src/asset_router.rs
```diff
@@ -908,10 +908,15 @@ impl<'content> AssetRouter<'content> {
...
+            // The `Range` request header will not be sent with the first request,
+            // so we don't include it in certification for the first chunk.
+            if range_begin != 0 {
+                request_headers.push((
+                    http::header::RANGE.to_string(),
+                    format!("bytes={range_begin}-"),
```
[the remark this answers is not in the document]
nathanosdev: At this point we are preparing the response certification inside the `init` or `post_upgrade` hooks, so there's no request to take the byte range from and we have to predict what byte ranges will be requested by the HTTP Gateway.

packages/ic-http-certification-tests/tests/v2_response_verification_sad_path.rs
```diff
@@ -275,16 +283,41 @@ mod tests {
...
-                (ResponseVerificationError::CertificateVerificationFailed(CertificateVerificationError::SignatureVerificationFailed), ResponseVerificationError::CertificateVerificationFailed(CertificateVerificationError::SignatureVerificationFailed)) |
-                (ResponseVerificationError::CertificateVerificationFailed(CertificateVerificationError::PrincipalOutOfRange { .. }), ResponseVerificationError::CertificateVerificationFailed(CertificateVerificationError::PrincipalOutOfRange { .. })) |
-                (ResponseVerificationError::CertificateVerificationFailed(CertificateVerificationError::TimeTooFarInThePast { .. }), ResponseVerificationError::CertificateVerificationFailed(CertificateVerificationError::TimeTooFarInThePast { .. })) |
-                (ResponseVerificationError::CertificateVerificationFailed(CertificateVerificationError::TimeTooFarInTheFuture { .. }), ResponseVerificationError::CertificateVerificationFailed(CertificateVerificationError::TimeTooFarInTheFuture { .. }))
-            ))
-        );
+        assert_matches!(
+            (result, expected_failure),
+            (
```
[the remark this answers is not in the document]
nathanosdev: We're matching two things here:
- That `result` and `expected_failure` are the same error.
- The `result` is one of the expected failures from the test case.

The tuple of `(result, expected_failure)` is matched against a tuple of possible errors like `(ErrorOne, ErrorOne)`, `(ErrorTwo, ErrorTwo)` so we make sure that the result matches one of the expected failures and also matches the current expected failure.

`assert_matches!(result, Err(expected_failure))` would only match that `result` is an `Err`, we'd need to extend it to be:
```
assert_matches!(result, Err(expected_failure) if expected_failure == result)
```
But that doesn't work because the errors don't implement `Eq` or `PartialEq`.

packages/ic-asset-certification/tests/large_assets.rs
```diff
@@ -0,0 +1,193 @@
...
+use common::*;
+
+const ASSET_ONE_NAME: &str = "asset_one";
+const ASSET_ONE_SIZE: usize = ASSET_CHUNK_SIZE + 1;
+
+const MAX_CERT_TIME_OFFSET_NS: u128 = 300_000_000_000;
+const MIN_REQUESTED_VERIFICATION_VERSION: u8 = 2;
+
+#[rstest]
+fn test_large_assets(
```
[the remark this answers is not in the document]
nathanosdev: Done, thanks

packages/ic-asset-certification/tests/large_assets.rs
```diff
@@ -0,0 +1,193 @@
...
+    let certified_data = asset_router.root_hash();
+    let CertificateData {
+        cbor_encoded_certificate,
+        certificate: _,
+        root_key,
+    } = CertificateBuilder::new(&canister_id.to_string(), &certified_data)
+        .unwrap()
+        .with_time(current_time)
+        .build()
+        .unwrap();
```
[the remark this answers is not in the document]
nathanosdev: Done, thanks!

packages/ic-asset-certification/src/asset_router.rs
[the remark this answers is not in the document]
przydatek: Ah, right, thanks for explaining!

packages/ic-http-certification-tests/tests/v2_response_verification_sad_path.rs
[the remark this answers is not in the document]
przydatek: I see, the test does check the right thing, but in a convoluted way...
It is unfortunate, especially that the big "or" on the pairs of errors has to be in sync with the cases of the test.  There should be a way to simplify that ... 🤔 ... but given that this is not the topic of this PR, it should not block it.

przydatek (approved):
LGTM, thanks!

PR #438, opened 2025-04-22
feat: use StatusCode instead of numbers in json-api impl example

peterpeterparker:
# Motivation

When I copied the code from the json-api example, I noticed that numbers were hardcoded for the status codes in the API response implementation. It's a bit opinionated, but since the example's codebase already uses `http::StatusCode`, I figured it made sense to do the same there as well.


raymondk:
This will need to be rebased after https://github.com/dfinity/response-verification/pull/439 is merged to get all the tests to pass...

nathanosdev (approved):

PR #493, opened 2026-09-11
chore(deps): bump anyhow, event-listener, rand

SafraNako:
## What

`Cargo.lock` pinned three crates below their patched versions:

| Crate | Bump | Advisory |
|---|---|---|
| `anyhow` | `1.0.100` → `1.0.103` | [RUSTSEC-2026-0190](https://rustsec.org/advisories/RUSTSEC-2026-0190.html) — unsoundness in `Error::downcast_mut()` |
| `event-listener` | `5.4.1` → `5.4.2` (a separate `2.5.3` instance elsewhere in the graph is a different major, unaffected) | [RUSTSEC-2026-0221](https://rustsec.org/advisories/RUSTSEC-2026-0221.html) — a `!Send` value could cross a thread boundary via `StackSlot` |
| `rand` | `0.8.5` → `0.8.6` (a separate `0.10.2` instance elsewhere is already patched) | [RUSTSEC-2026-0097](https://rustsec.org/advisories/RUSTSEC-2026-0097.html) / [GHSA-cq8v-f236-94qc](https://github.com/advisories/GHSA-cq8v-f236-94qc) — unsound with a custom logger using `rand::rng()` |

`Cargo.toml` ranges already allow all three, so this is `cargo update -p
<crate> --precise <version>` only — no manifest or source changes.

## Verified locally

`cargo test --workspace --lib --bins`, excluding the example canister
backends (they need a pre-built `frontend/dist` I don't have available) and
`ic-response-verification-wasm`: everything else passes — 56 real tests plus
several zero-test crates compiling cleanly.

## Unrelated finding, not fixed here

`ic-response-verification-wasm` fails to compile on `main` as-is (verified
via `git stash` before making this change, so it's not something this PR
introduces): `ic_response_verification::ResponseVerificationJsError` no
longer exists and `JsValue: From<VerificationInfo>` isn't implemented.
Flagging in case it isn't already known — happy to look into it separately
if useful.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


raymondk:
Thanks @SafraNako - the compilation failure is a known issue. That package is not meant to be built for the host only for wasm32 - see https://github.com/dfinity/response-verification/blob/main/Cargo.toml#L26

raymondk (approved):

